Issue #7, opened 2015-01-18
MacOS Yosemite support


m3m0ryh0l3:
When running on MacOS Yosemite (10.10.1), it fails randomly in malloc (usually in qt_affinity_gendists) complaining about memory corruption if and only if QT_NUM_SHEPHERDS is > 6 (whether set manually, or determined by sysctl()).

olivier-snl:
I just merged labnerd's pull request which I think should have resolved this problem.

PR #17, opened 2016-02-18
Bugfix and testing file

labnerd:
These changes contain a bug fix for #7 and a new yml file for GitLab automated testing support.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,6 @@
+
+runTests:
+    script:
+        - ./autogen.sh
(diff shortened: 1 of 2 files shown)


olivier-snl:
Note for documentation purposes: The type of qthread_shep_dist is unsigned int, and the qthread_shepherd_id_t type is unsigned short, so this fixes the calloc to qthread_shep_dist to be unsigned int.  Guessing that this problem arose in the first place due to a copy and paste error from the line above, where the calloc using type qthread_shepherd_id_t is correct.

Issue #3, opened 2015-01-18
New qt_dictionary_iterator does not point to first element


dylan-stark:
# Reported by ndryden@ndryden.com, Sep 3, 2014

**What steps will reproduce the problem?**

```
qt_dictionary* dict = qt_dictionary_create(/* ... */);
/* Insert some stuff. */
qt_dictionary_iterator* iter = qt_dictionary_iterator_create(dict);
list_entry* e = qt_dictionary_iterator_get(iter);
assert(e);
```

**What is the expected output? What do you see instead?**

qt_dictionary_iterator_get should return the first element of the dictionary. Instead, it returns NULL.

**What version of the product are you using? On what operating system?**

Qthreads 1.10 on x86-64 Linux.

**Please provide any additional information below.**

Might just be the documentation (http://www.cs.sandia.gov/qthreads/man/qt_dictionary_iterator_create.html) is out of date.eeeeee

npe9:
I'm not sure this is the expected behavior. Looking at the tests I see the invocation:

` 
qt_dictionary_iterator *it         = qt_dictionary_iterator_create(dict);
    while(NULL != qt_dictionary_iterator_next(it)) {
        list_entry *le = qt_dictionary_iterator_get(it);
        no_entries++;
        assert(le != NULL);
        iprintf("17. Pair:(%s,%s)\n", (char *)le->key, (char *)le->value);
    }
`

which means that an initial value of NULL is the expected semantics, since qt_dictionary_iterator_next(), not the iterator itself creates the values in the iterator.

PR #19, opened 2016-03-01
Add issues test directory to allow tests to be created from the issue…

npe9:
…s on github. Populate it with a Makefile.am and the two outstanding issues with test code.


npe9:
This failure was on purpose, see my comment about issue #3

Issue #22, opened 2016-05-26
Missing some FEB operations that XMT supported


ronawho:
I'm in the process of trying to map Chapel sync vars down to
qthreads aligned_t FEBs. I was under the impression that qthreads
FEBs were heavily based off the XMT/MTA, so I was surprised that
some of the operations that the XMT supported aren't available with
qthreads. In particular writeFF, readXX, and purge/reset are
missing.

I'm curious if anybody knows the reason that these functions were
omitted and if you'd be willing to consider adding support for them.
I realize that it's pretty unlikely that anybody currently on the
qthreads team would know the history off the top of their head.
Maybe @m3m0ryh0l3 would know?

The table below outlines the FEB operations that Chapel, XMT, and
Qthreads support. I *might* be able to implement readXX and purge by
composing other functions since they don't interact with the F/E
state, but writeFF would require Qthreads support

| Chapel  | XMT     | Qthreads           |
| ------- | ------- | ------------------ |
| readFE  | readfe  | qthread_readFE     |
| readFF  | readff  | qthread_readFF     |
| readXX  | readxx  |                    |
| writeEF | writeef | qthread_writeEF    |
| writeFF | writeff |                    |
| writeXF | writexf | qthread_writeF     |
| reset   | purge   |                    |
| isFull  |         | qthread_feb_status |
|         |         | qthread_empty      |
|         |         | qthread_fill       |

ronawho:
If it helps, I took a look at what it would take to implement writeFF:

I think I have a version that works. I basically copied writeEF and modified a
few sections.

My changes are against the qthreads version in the chapel repo:

https://github.com/chapel-lang/chapel/compare/master...ronawho:add-qthreads-writeff?expand=1

That diff looks big, but I broke it into 2 logical commits. One that just
copies writeEF as writeFF, and a second that makes the semantic changes needed.
The second commit is actually pretty small:

https://github.com/chapel-lang/chapel/commit/cc480c7

There's obviously more that would need to be done for this, but I think it's a
good proof of concept that adding writeFF might not be all that hard

olivier-snl:
[...]

PR #24, opened 2016-08-09
Add aligned writeff

ronawho:
This adds support for writeFF on aligned (but not syncvar_t) FEB vars

In order to support writeFF I basically:
 - added a new FEB queue called FFWQ (full full writers (aka writeFF) queue)
 - cloned readFF to writeFF since the hash locking/unlocking and when we can
   read/write is similar
 - updated writeFF to
    - use a WRITEFF blocker_type
    - use dest instead of src for lockbin, alignedaddr, and x->addr
    - use FFWQ instead of FFQ

Note that when an empty FEB is filled, the current implemenation performs all
writeFF operations first, and then continues onto the existing readFF
operations.

This also adds some writeFF tests to check that:
 - writeFF on a full var writes, and leaves the FEB full
 - concurrent writeFFs work without tearing
 - writeFF on an empty var blocks until filled, then writes, and leaves full

Note that I did not add writeFF for the syncvar_t FEB type, and I don't have
any plans to do so. This is being added because we want to map Chapel sync vars
down to native qthread syncvars and we need the full 64 bits so we can't use
the inline FEBs.

writeE and readXX are still needed in order to map Chapel Sync vars down. I
plan on adding those shortly, but I split writeFF off to simplify review.
https://github.com/Qthreads/qthreads/issues/22 has more info on the missing FEB
operations.

### DIFF
--- a/include/qt_addrstat.h
+++ b/include/qt_addrstat.h
@@ -17,6 +17,7 @@ static QINLINE qthread_addrstat_t *qthread_addrstat_new(void)
         ret->EFQ   = NULL;
         ret->FEQ   = NULL;
         ret->FFQ   = NULL;
+        ret->FFWQ  = NULL;
         QTHREAD_EMPTY_TIMER_INIT(ret);
     }
     return ret;
--- a/include/qt_blocking_structs.h
+++ b/include/qt_blocking_structs.h
@@ -49,6 +49,7 @@ typedef struct qthread_addrstat_s {
     qthread_addrres_t    *EFQ;
     qthread_addrres_t    *FEQ;
     qthread_addrres_t    *FFQ;
+    qthread_addrres_t    *FFWQ;
 #ifdef QTHREAD_FEB_PROFILING
     qtimer_t              empty_timer;
 #endif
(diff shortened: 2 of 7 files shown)


ronawho:
Passes `make check` on osx with gcc/clang, linux64 with gcc, and on a cray machine with gnu, intel, cray, and pgi compilers

stelleg:
Looks great, thanks!

m3m0ryh0l3:
What is the use-case of writeFF?

ronawho:
I'm not sure what good use-cases there are for `writeFF`. Out of ~7,000 tests, we only have 2 or 3 uses of it, and they seem to just setup initial values for sync arrays

Something like:

```chapel
var visited$: [1..10] sync int = -1;
visited$[0].writeFF(1);
```

and so far as I can tell all these tests could just as easily use `writeF` instead.

I'll check with Greg and Sung or maybe some XMT people that I sit close to, and see if I can come up with a real use-case. If your question is alluding to "why bother supporting writeff", I don't currently have a good answer beyond Chapel currently has `writeFF` and I needed something to map down to.

ronawho:
This isn't much more compelling, but another possible use-case that Brad suggested is something like:

```chapel
/* outside a parallel region */
var val = mySync$.readFF();
val = foo(val);
mySync$.writeFF(val);
```

although here again `writeF` *could* be used in place of `writeFF`. Nobody else could think of examples off the top of their heads and some of the XMT people wondered if it was just added for "completeness". I'll let you know if they dredge up any uses in old code they have lying around, but even if they don't, at this point Chapel isn't going to drop `writeFF` without a compelling reason.

PR #25, opened 2016-08-19
Add aligned writeE

ronawho:
This adds support for writeE on aligned (but not syncvar_t) FEB vars

In order to support writeE I basically cloned qthread_empty since the hash
locking/unlocking is the same, and added a write after the locking logic.

This also adds some writeE tests to check that:
 - writeE on a full var writes, and empties the FEB
 - writeE on an empty var writes, and leaves the FEB empty
 - writeE on a full var wakes a blocked writeEF

### DIFF
--- a/include/qthread/qthread.h
+++ b/include/qthread/qthread.h
@@ -487,6 +487,21 @@ int qthread_syncvar_writeF(syncvar_t *restrict      dest,
 int qthread_syncvar_writeF_const(syncvar_t *restrict dest,
                                  uint64_t            src);
 
+/* This function is a cross between qthread_empty() and qthread_writeEF(). It
+ * does not wait for memory to become empty, but performs the write and sets
+ * the state to empty atomically with respect to other FEB-based actions. Data
+ * is read from src and written to dest.
+ *
+ * The semantics of writeE are:
+ * 1 - data is copied from src to destination
+ * 2 - the destination's FEB state gets set to empty
+ */
+int qthread_writeE(aligned_t *restrict       dest,
+                   const aligned_t *restrict src);
+int qthread_writeE_const(aligned_t *dest,
+                         aligned_t  src);
+// NOTE: There is no syncvar version of writeE or writeE_const
+
 /* This function waits for memory to become full, and leaves it full. When
  * memory becomes full, all threads waiting for it to become full with a
  * writeFF will write their value and be queued to run. Data is read from src
--- a/src/feb.c
+++ b/src/feb.c
@@ -46,6 +46,7 @@ QTHREAD_FASTLOCK_TYPE *febs_stripes_locks;
  * Local Types
  *********************************************************************/
 typedef enum bt {
+    WRITEE,
     WRITEEF,
     WRITEEF_NB,
     WRITEF,
@@ -204,6 +205,9 @@ static aligned_t qthread_feb_blocker_thread(void *arg)
             break;
         case READFF_NB:
             a->retval = qthread_readFF_nb(a->a, a->b);
+            break;
+        case WRITEE:
+            a->retval = qthread_writeE(a->a, a->b);
             break;
         case WRITEEF:
             a->retval = qthread_writeEF(a->a, a->b);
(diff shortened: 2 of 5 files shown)


ronawho:
Passes `make check` on osx with gcc/clang, linux64 with gcc, and on a cray machine with gnu, intel, and cray compilers. 

I'm pretty confident this code is ok. The part that I'm not 100% sure about is if it's ok that for the case where the FEB is full and there's no pending operations (the path where we create the addrstat) this doesn't lock the addrstat, so we end up performing the write without a lock. This is true for writeF and writeFF too, which is why I think it's fine, but again I'm not 100% sure.

stelleg:
Thanks!

Looks good, code's almost identical to qthead_empty with the addition of the write. Should be fine and passing test cases.

m3m0ryh0l3:
It's already merged, so perhaps there's no point in commenting, but... this makes it sound like full/empty state is, essentially meaningless. If you can write into a value and that write causes it to become empty - and you can read that value out of the "empty" state, then what on earth does it mean to be "empty"? It may as well be called red/blue synchronization, if you're just going to have a full complement of read/write operations that can wait for either state to be achieved (or, optionally, not wait) and atomically set the state to any other state.

ronawho:
Here, `writeE` is really meant to be `purge`/`reset` (Note that in contrast to qthreads, uninitialized sync vars on XMT and in Chapel are empty)

```chapel
var s$ : sync int;       // empty
var s1$: sync int = -1   // full
```

For [XMT](http://user.cscs.ch/fileadmin/user_upload/customers/CSCS_Application_Data/Files/Presentations/Courses_Ws_2011/CRAY_XMT_Programming_Jun11/S-2479-20.pdf) `purge` "Writes 0, using the appropriate data type, to variable v and sets v to an empty state."

and for [Chapel](http://chapel.cray.com/docs/latest/_downloads/chapelLanguageSpec.pdf) `reset` "Assigns the default value of type t to the value of the sync variable..."

In Chapel, a sufficiently motivated user can override the default value of a type, so instead of `int` defaulting to 0, it could default to 123 or whatever. I can't imagine why anybody would do that, but it's possible, so writeE is really meant to be purge/reset, but it takes src so that we can specify a default value to purge to.


None of that addresses your question of whether an empty variable has a determinate value though. If it doesn't, then purge could just be implemented using qthread_empty since the value would be undefined anyways. My understanding is that Chapel expects an empty var to maintain the value it had, and that for example readXX on a variable that was just readFE'd should have the same value as when it was full.

Neither our spec nor the XMT manual explicitly say one way other the other though, but many of our tests certainly expect that behavior, and it seems odd that XMT would have readXX if reading an empty variable wasn't expected to produce some valid result.

I'll add "what does readXX on an empty var do?" to the list of questions I need to ask Greg/Sung/XMT folks. If the answer is undefined behavior, then we can get rid of writeE and just use qthread_empty, but I would guess that we expect emptied vars to maintain their value.

ronawho:
>None of that addresses your question of whether an empty variable has a determinate value though. If it doesn't, then purge could just be implemented using qthread_empty since the value would be undefined anyways. My understanding is that Chapel expects an empty var to maintain the value it had, and that for example readXX on a variable that was just readFE'd should have the same value as when it was full.

There's definitely consensus  that an empty variable should maintain its value from Greg/Sung/Brad and the XMT folks.

Stealing a quote from one of them: "I see Kyle's point as well from a purist point of view, but I think following the XMT's lead until proven wrong is probably reasonable and am not particularly interested in reinventing this aspect of Chapel without a compelling rationale."

And note that it seems mtgl expected empty sync vars to maintain a state too. Purge was implemented as:

```c++
template <typename T>
void mt_purge(T& target)
{
#ifdef __MTA__
  purge(&target);
#elif USING_QTHREADS
  qthread_empty(&target);
  target = 0;
#else
  target = 0;
#endif
}
```

and they had a readXX that just did  naked read for qthreads:

```c++
template <typename T>
T mt_readxx(T& target)
{
#ifdef __MTA__
  return readxx(&target);
#else
  return target;
#endif
}
```

So in order to map Chapel sync vars down to native qthreads ones, we need empty sync vars to maintain their value (and looking at uses in mtgl, that seems to be the case for them too.)

Does that alleviate your concerns about `writeE` at all? Would it be better if it was called `purge` or `reset`?

m3m0ryh0l3:
Personally, I think I prefer calling it something like "reset", just aesthetically. However, the question then becomes, how does this interact with a large queue of pending operations? A "reset" or "purge" operation might reasonably be expected to discard any pending operations, and has only one value it can write (0, presumably), whereas "writeE" is just another arbitrary operation in the queue. But then, I find the whole concept of using a write to empty variables weird - it seems like a violation of the established semantics of full/empty, so maybe I'm not the best one to be rendering an opinion.

ronawho:
I'm game for changing to reset or purge.

And it would be programmer error to use reset while other sync var operations are going on in parallel. Just like doing a writeF in the middle of pending operations could be

Stealing from the XMT docs again:

```
However, if any generic function in the set does not depend on the
full-empty state (such as the purge, readxx, and writexf functions), the
ability to serialize the set is not guaranteed. If the set is not
serialized, generic functions may access words in the variable in a
different order, resulting in inconsistencies in one or more of the
following: the state of the value returned by one or more of the generics;
the memory holding the variable; the data value; or the full-empty bits
```

I believe purge/reset is really just meant as a way to reset the sync variable back to it's initial state in a non parallel region

ronawho:
`qthread_writeE` renamed to `qthread_purge_to` with https://github.com/Qthreads/qthreads/pull/32. That also adds a `qthread_purge`, which is intended to be the "native" qthreads version, and the `purge_to` version is intended only for use by Chapel

PR #26, opened 2016-08-19
Add aligned readXX

ronawho:
This adds support for readXX on aligned (but not syncvar_t) FEB vars

qthread_readXX is really just a simple wrapper around:

```c
if (dest && (dest != src)) {
    *(aligned_t *)dest = *(aligned_t *)src;
}
```

### DIFF
--- a/src/feb.c
+++ b/src/feb.c
@@ -54,7 +54,6 @@ typedef enum bt {
     READFF,
     READFF_NB,
     READFE,
-    READXX,
     READFE_NB,
     FILL,
     EMPTY
@@ -206,9 +205,6 @@ static aligned_t qthread_feb_blocker_thread(void *arg)
             break;
         case READFF_NB:
             a->retval = qthread_readFF_nb(a->a, a->b);
-            break;
-        case READXX:
-            a->retval = qthread_readXX(a->a, a->b);
             break;
         case WRITEE:
             a->retval = qthread_writeE(a->a, a->b);
@@ -1476,19 +1472,10 @@ got_m:
 int API_FUNC qthread_readXX(aligned_t *restrict       dest,
                             const aligned_t *restrict src)
 {                      /*{{{ */
-    qthread_t          *me      = qthread_internal_self();
-
-    assert(qthread_library_initialized);
-
-    if (!me) {
-        return qthread_feb_blocker_func(dest, (void *)src, READXX);
-    }
-    qthread_debug(FEB_CALLS, "dest=%p, src=%p (tid=%i)\n", dest, src, me->thread_id);
-    QTHREAD_FEB_UNIQUERECORD(feb, src, me);
+    qthread_debug(FEB_CALLS, "dest=%p, src=%p\n", dest, src);
 
     if (dest && (dest != src)) {
         *(aligned_t *)dest = *(aligned_t *)src;
-        MACHINE_FENCE;
     }
     return QTHREAD_SUCCESS;
 }                      /*}}} */
--- a/src/feb.c
+++ b/src/feb.c
@@ -54,6 +54,7 @@ typedef enum bt {
     READFF,
     READFF_NB,
     READFE,
+    READXX,
     READFE_NB,
     FILL,
     EMPTY
@@ -205,6 +206,9 @@ static aligned_t qthread_feb_blocker_thread(void *arg)
             break;
         case READFF_NB:
             a->retval = qthread_readFF_nb(a->a, a->b);
+            break;
+        case READXX:
+            a->retval = qthread_readXX(a->a, a->b);
             break;
         case WRITEE:
             a->retval = qthread_writeE(a->a, a->b);
@@ -1461,6 +1465,31 @@ got_m:
         qthread_gotlock_empty(me->rdata->shepherd_ptr, m, (void *)alignedaddr);
     }
     QTHREAD_FEB_TIMER_STOP(febblock, me);
+    return QTHREAD_SUCCESS;
+}                      /*}}} */
+
+/* the way this works is that:
+ * 1 - src's FEB state is ignored
+ * 2 - data is copied from src to destination
+ */
+
+int API_FUNC qthread_readXX(aligned_t *restrict       dest,
+                            const aligned_t *restrict src)
+{                      /*{{{ */
+    qthread_t          *me      = qthread_internal_self();
+
+    assert(qthread_library_initialized);
+
+    if (!me) {
+        return qthread_feb_blocker_func(dest, (void *)src, READXX);
+    }
+    qthread_debug(FEB_CALLS, "dest=%p, src=%p (tid=%i)\n", dest, src, me->thread_id);
+    QTHREAD_FEB_UNIQUERECORD(feb, src, me);
+
+    if (dest && (dest != src)) {
+        *(aligned_t *)dest = *(aligned_t *)src;
+        MACHINE_FENCE;
+    }
     return QTHREAD_SUCCESS;
 }                      /*}}} */
 
--- a/test/basics/aligned_readXX_basic.c
+++ b/test/basics/aligned_readXX_basic.c
@@ -1,0 +1,60 @@
+#ifdef HAVE_CONFIG_H
+# include <config.h>
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <unistd.h>
+#include <qthread/qthread.h>
+#include "argparsing.h"
+
+// Test that a readXX on a full var performs the read, and leaves the FEB state
+// unchanged
+static void testReadXXOnFull(void)
+{
+    aligned_t x, val;
+
+    x = 45, val = 55;
+    assert(qthread_feb_status(&x) == 1);
+
+    iprintf("Before x=%d, val=%d, x_full=%d\n", x, val, qthread_feb_status(&x));
+    qthread_readXX(&val, &x);
+    iprintf("After  x=%d, val=%d, x_full=%d\n", x, val, qthread_feb_status(&x));
+
+    assert(qthread_feb_status(&x) == 1);
+    assert(x == 45);
+    assert(val == 45);
+}
+
+// Test that a readXX on an empty var performs the read, and leaves the FEB
+// state unchanged
+static void testReadXXOnEmpty(void)
+{
+    aligned_t x, val;
+
+    x = 45, val = 55;
+    qthread_empty(&x);
+
+    iprintf("Before x=%d, val=%d, x_full=%d\n", x, val, qthread_feb_status(&x));
+    qthread_readXX(&val, &x);
+    iprintf("After  x=%d, val=%d, x_full=%d\n", x, val, qthread_feb_status(&x));
+
+    assert(qthread_feb_status(&x) == 0);
+    assert(x == 45);
+    assert(val == 45);
+}
+
+int main(int argc,
+         char *argv[])
+{
+    CHECK_VERBOSE();
+    assert(qthread_initialize() == 0);
+    iprintf("%i shepherds...\n", qthread_num_shepherds());
+    iprintf("  %i threads total\n", qthread_num_workers());
+
+    testReadXXOnFull();
+    testReadXXOnEmpty();
+}
+
+/* vim:set expandtab */
--- a/include/qthread/qthread.h
+++ b/include/qthread/qthread.h
@@ -546,6 +546,17 @@ int qthread_readFE(aligned_t       *dest,
 int qthread_syncvar_readFE(uint64_t *restrict  dest,
                            syncvar_t *restrict src);
 
+/* This function ignores the FEB state. Data is read from src and written to
+ * dest.
+ *
+ * The semantics of readXX are:
+ * 1 - src's FEB state is ignored
+ * 2 - data is copied from src to destination
+ */
+int qthread_readXX(aligned_t       *dest,
+                   const aligned_t *src);
+// NOTE: There is no syncvar version of readXX
+
 /* functions to implement FEB-ish locking/unlocking
  *
  * These are atomic and functional, but do not have the same semantics as full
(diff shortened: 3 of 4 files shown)


ronawho:
Passes `make check` on osx with gcc/clang, linux64 with gcc, and on a cray machine with gnu, intel, and cray compilers. 

I think because readXX doesn't interact with the FEB state, it's fine to just do a read without locking the hash or an addrstat, but maybe I'm missing something and it should look something more like [qthread_feb_status](https://github.com/Qthreads/qthreads/blob/master/src/feb.c#L254-L298)

ronawho:
Oh, and I'm pretty sure this will have merge conflicts with #25, so I'll rebase after that gets committed

stelleg:
Code looks good, I'll merge after rebase.

ronawho:
rebased

m3m0ryh0l3:
A question: what is the correct value of an empty variable? Does it even have a correct value? I ask because there are some who think empty variables can have their storage re-used to contain pointers to management data structures, and the question becomes: would that break something? Is there a reason to expect that readXX will return a value that should be trusted in any way? I realize that, much like uninitialized memory, yes, it has a value, but... reading uninitialized memory is often considered a programming error (and many compilers will warn you loudly). Isn't readXX, on a potentially empty value, essentially a racy read on potentially uninitialized data?

Another question: Does readXX have any ordering semantics? Does it prevent compiler re-ordering around the read? Can the readXX be hoisted to other parts of the program (following all the same rules that a basic read operation would)? If there's a currently executing sequence (backlog) of readFE/writeEF operations, does readXX guarantee any particular value?

I guess what I'm getting at is: why would a programmer use readXX? Why does Chapel include readXX? And what semantics (if any) does it have? Its existence suggests a polling-style of programming that is antithetical to the purpose of fine-grained synchronization in the first place.

src/feb.c
```diff
@@ -1464,6 +1468,31 @@ int API_FUNC qthread_readFE(aligned_t *restrict       dest,
...
+
+int API_FUNC qthread_readXX(aligned_t *restrict       dest,
+                            const aligned_t *restrict src)
+{                      /*{{{ */
+    qthread_t          *me      = qthread_internal_self();
+
+    assert(qthread_library_initialized);
+
+    if (!me) {
+        return qthread_feb_blocker_func(dest, (void *)src, READXX);
```
m3m0ryh0l3: This operation explicitly cannot block, so why use the blocker_func? This seems wrong.

src/feb.c
```diff
@@ -1464,6 +1468,31 @@ int API_FUNC qthread_readFE(aligned_t *restrict       dest,
...
+    }
+    qthread_debug(FEB_CALLS, "dest=%p, src=%p (tid=%i)\n", dest, src, me->thread_id);
+    QTHREAD_FEB_UNIQUERECORD(feb, src, me);
+
+    if (dest && (dest != src)) {
+        *(aligned_t *)dest = *(aligned_t *)src;
+        MACHINE_FENCE;
```
m3m0ryh0l3: Why is there a `MACHINE_FENCE` here? Why not a `MACHINE_FENCE `*before* the copy operation?

I think this gets back to the question of what valid values for readXX are when the value you're reading is (or could be) empty. (see my comment on https://github.com/Qthreads/qthreads/issues/22)

test/basics/aligned_readXX_basic.c
```diff
@@ -0,0 +1,60 @@
...
+    x = 45, val = 55;
+    qthread_empty(&x);
+
+    iprintf("Before x=%d, val=%d, x_full=%d\n", x, val, qthread_feb_status(&x));
+    qthread_readXX(&val, &x);
+    iprintf("After  x=%d, val=%d, x_full=%d\n", x, val, qthread_feb_status(&x));
+
+    assert(qthread_feb_status(&x) == 0);
+    assert(x == 45);
+    assert(val == 45);
```
m3m0ryh0l3: So, this would seem to suggest that empty variables MUST have the same value they had when they were full, yes? Why?

src/feb.c
ronawho: The only reason I had it use the blocker_func was so that it was guaranteed to run in qthread, otherwise the shepherd passed to qthread_debug and UNIQUERECORD could be null. 

Maybe that's not a good enough reason, in which case it's easy to remove from the debug call, and maybe we don't need to call UNIQUERECORD for readXX anyways

m3m0ryh0l3: Certainly, you don't need `QTHREAD_FEB_UNIQUERECORD` - that's only there to profile the performance of the central striped hash table (when `FEB_PROFILING` is turned on) - if you don't use that hash table, you shouldn't use `QTHREAD_FEB_UNIQUERECORD` (in my opinion).

As for qthread_debug, in my personal opinion, debugging output should *never* require heavyweight operations - especially not ones that get triggered regardless of whether debugging is turned on or not. I prefer my code lean and mean when debugging isn't on.

ronawho: I considered #ifdefing the blocker func call if debugging or FEB profiling was turned on, but decided against it, because in the presence of other FEB operations, doing a blocker func seemed relatively cheap. i.e. who cares if readxx is a little more expensive considering the cost of other FEB operations that are likely going to occur in the same program

But I think you're right that if we don't need `QTHREAD_FEB_UNIQUERECORD`, then there's no reason to include the shep id in the debug code, so I'll just remove all that

ronawho:
readXX (and all the operations that don't depend on the FEB such as writeXF, reset, etc) can be reordered and there is no guarantee about the value of readXX amongst a backlog of readFE/writeEF operations. More formal version stolen from the XMT manual:

    However, if any generic function in the set does not depend on the
    full-empty state (such as the purge, readxx, and writexf functions), the
    ability to serialize the set is not guaranteed. If the set is not
    serialized, generic functions may access words in the variable in a
    different order, resulting in inconsistencies in one or more of the
    following: the state of the value returned by one or more of the generics;
    the memory holding the variable; the data value; or the full-empty bits


Most of our uses for readXX are for debugging or just printing out the value in correctness tests without thinking about the FEB state.

The real use-case that I see for readXX is as a fast pre-test to avoid doing a full readFE when possible. Below example stolen from `InsertLinearProbing` in  http://cass-mt.pnnl.gov/docs/pubs/hashingstrategies.pdf  

```fortran
// Procedure: InsertLinearProbing
// Parameters: (lock, slotKey, candidateKey)

status = readxx(lock)
if status == claimed then
  if slotKey == candidatekey then
    return true
  end if
else
  status = readfe(lock)
  // The first thread to reach the slot?
  if status == unclaimed then
    slotKey = candidateKey
    writeef(lock, claimed)
    return true
  else
    if slotKey == candidateKey then
      writeef(lock, claimed)
      return true
    end if
    writeef(lock, claimed)
  end if
  return false
end if

```

We have some code that does something similar,  which I think indicates that empty variables maintain their values. Or.. it indicates that we and that paper's authors are misusingreadXX

ronawho:
And FWIW the only reason I added qthread_readXX instead of just doing the read directly was because I was thinking that a direct load of a 64-bit aligned_t wouldn't be atomic on a 32-bit processor, so you would need readXX to make the double word load atomic somehow.

Chapel's `int` is 64-bit regardless of the platform, so I was going to build qthreads with `--enable-64bit-aligned_t` but then I got to thinking that the unprotected loads/stores wouldn't actually be atomic on a 32-bit platform and that things would probably break.

But maybe the real question with that is who cares about what happens on 32-bit processors, the only ones we're likely to run into nowadays are on embedded systems and neither chapel nor qthreads are currently optimized for that anyways.

stelleg:
Thanks @m3m0ryh0l3 ! Really helpful to get input from the person who actually knows what all the FEB code is doing. If you see any issues with commits that are already merged that I may have missed please feel free to comment on those merges and we can try and get those fixed as soon as possible. In particular I'm a little concerned with your mention of using memory in an empty addrstat for bookkeeping, and whether or not writeE could break that, as I've run into that issue with TLS for stacks. It doesn't look like that could happen from the code, but I could have missed something.

I think one of the overarching concerns is that the semantics of some of these new functions is not obvious. I think for this reason, and because we have them for the existing FEB operations, we should include with these PRs another PR with man pages that describe the relevant semantics in detail. Elliot, would you mind adding these in a separate PR? Your comments help, but I think more detailed semantics in the man pages would be more helpful, and like @m3m0ryh0l3 I'm unsure exactly what the intended semantics are.

test/basics/aligned_readXX_basic.c
stelleg: I agree that this seems a little misleading given the semantics of readXX. Perhaps @ronawho you could remove the assert and make it clear in comments that while this will succeed in this case it can't be guaranteed in general?

src/feb.c
stelleg: I'd suggest either @m3m0ryh0l3's suggestion or removing the `MACHINE_FENCE` altogether given that it's purpose is to be fast and that it has no real guarantees about what it returns.

m3m0ryh0l3: +1 to removing the `MACHINE_FENCE` altogether.

m3m0ryh0l3:
@ronawho - I certainly understand and appreciate the thought that readXX wouldn't be atomic. I agree with you that it's unlikely to be important - 64-bit is pretty much the primary target of Chapel, but that's the kind of thing that ought to be made formal (specifically, that Chapel _does NOT_ support 32-bit platforms). If it can't be made formal, then this readXX implementation isn't good enough, because just the naked read isn't atomic, and we'd have to do some real thinking about the best way to make it atomic. Of course, that leads us down the path of trying to define (and then rely upon) the Chapel memory model, and whether or not regular 64-bit int reads are expected to be atomic (or 128-bit int reads, or unaligned 17-bit reads, or whatever). I'm sure there's plenty of hard thought that needs to be done on the subject.

On the other hand, if readXX doesn't provide a guarantee about the value stored there - and that's what the spec you quoted sounds like - then tearing/corrupt data is probably the least of our worries. Frankly, readXX sounds like its value cannot be trusted without some other memory-model-level synchronization that would ensure a specific value. It might be fast, but as specified, it doesn't sound reliable.

@stelleg Glad to be helpful - this is the fun stuff! :) Anyway, the Qthread FEB semantics here are pretty arbitrary - they're whatever happened to be the simplest to implement. Specifically, when transitioning from empty to full, all the readFFs trigger, and then ONE of the readFEs triggers (the oldest). In other words, Qthreads doesn't maintain ordering between readFFs and readFEs. You could maintain ordering, but that would require a bit of a refactor on the bookkeeping.

src/feb.c
ronawho: removed

src/feb.c
ronawho: removed `MACHINE_FENCE`

ronawho:
> @ronawho - I certainly understand and appreciate the thought that readXX wouldn't be atomic. I agree with you that it's unlikely to be important - 64-bit is pretty much the primary target of Chapel, but that's the kind of thing that ought to be made formal (specifically, that Chapel does NOT support 32-bit platforms). If it can't be made formal, then this readXX implementation isn't good enough, because just the naked read isn't atomic, and we'd have to do some real thinking about the best way to make it atomic. Of course, that leads us down the path of trying to define (and then rely upon) the Chapel memory model, and whether or not regular 64-bit int reads are expected to be atomic (or 128-bit int reads, or unaligned 17-bit reads, or whatever). I'm sure there's plenty of hard thought that needs to be done on the subject.

I should have been more clear here. Chapel __does__ support 32-bit platforms, I just meant that our performance might be less than ideal for things like 64-bit sync operations. What I'm currently doing for mapping down to native qthread sync vars is that if `sizeof(sync type) < sizeof(void*)` then I default to our old style sync vars. I added readXX in case qthreads ever supports 64-bit sync vars on 32-bit platforms. By "who cares about what happens on 32-bit processors", all I meant was that there's no real motivation to add 64-bit sync var support on 32-bit platforms to qthreads because nobody really cares about the performance there, so maybe there was no good reason to add a readXX wrapper anyways.

>@stelleg Glad to be helpful - this is the fun stuff! :)

You've definitely been helpful! I know just enough about sync var semantics/implementation to be dangerous, but clearly not much more than that :)

test/basics/aligned_readXX_basic.c
ronawho: I just updated https://github.com/Qthreads/qthreads/pull/25 with the details, but Chapel expects/needs empty variables to have the same value they had when they were full. Assuming qthreads is also willing to do that too, I think this assert should stay, otherwise we probably can't use native qthread sync vars.

And note that the current qthreads implementation does preserve the value, or rather it doesn't go out of it's way anywhere to invalidate it.

ronawho:
@stelleg 

>I think one of the overarching concerns is that the semantics of some of these new functions is not obvious. I think for this reason, and because we have them for the existing FEB operations, we should include with these PRs another PR with man pages that describe the relevant semantics in detail. @ronawho, would you mind adding these in a separate PR? Your comments help, but I think more detailed semantics in the man pages would be more helpful, and like @m3m0ryh0l3 I'm unsure exactly what the intended semantics are.

I'm not strongly opposed to adding man pages, but I didn't in the first place because these operations are likely to only be used by Chapel, and more importantly because they're only availble for the `aligned_t`  version and not the `syncvar_t` variant. 

IMO to be part of the "official" API, you'd probably want these operations implemented for both sync var types, and I'm not willing to add them for `syncvar_t`. 

If you still want man pages to be added, I think I need to know if qthreads is willing to commit to empty variables maintaining their value, and if `writeE` should be renamed to `purge`

stelleg:
I agree that these extra operations will likely only be used by chapel. I'm fine having it still called writeE, and I can add some documentation to make it clear that it's supported for legacy purposes and not recommended for general use.  I'll add those changes soon, but in the meantime I'll go ahead and merge this commit. I'll leave issue #22 open until I've made the documentation changes.

PR #30, opened 2016-09-01
Destroy qtimer_t in sleep, usleep, and nanosleep

ronawho:
They were being leaked previously, so destroy them before returning


stelleg:
Whoops! Thanks!

PR #31, opened 2016-09-01
Make sure main() explicitly returns 0 in recently added FEB tests

ronawho:
Whoops, not all compilers have an implicit "return 0" if the end of main is
reached. These tests were failing with some versions of pgi because of this.


ronawho:
@stelleg

stelleg:
Good catch, thanks.

PR #32, opened 2016-09-02
Rename qthread_writeE to qthread_purge_to

ronawho:
writeE was really meant to implement purge, but because a sufficiently
motivated Chapel user can override the default value of types, we need to be
able to purge to a specific value instead of always purging to 0. However,
writeE was a bad name, which could lead users to think it's intended for
something more than just "resetting" a sync variable.

This changes qthread_writeE to qthread_purge_to and adds a qthread_purge, which
purges to 0. qthread_purge is the "public" version that native qthreads users would 
use, and the purge_to variant is intended only for Chapel.


stelleg:
This is for cray by cray, so fine by me.

PR #55, opened 2017-06-09
Fix ignored attributes warning

dmk42:
This fixes a warning encountered in Chapel, which is turned into an error when \-Werror is used.

Qthreads has a clever implementation of `static_assert()` called `QTHREAD_STATIC_ASSERT()` that works even with older versions of the C\+\+ standard.  Unfortunately, the stricter checking performed by gcc 7.1 causes `QTHREAD_STATIC_ASSERT()` to provoke a warning that attributes are being ignored in a template argument.

The fix proposed here is to use the standard `static_assert()` when it is available (i.e. when using C\+\+11 or later).

@ronawho suggested that I submit this as a Qthreads PR since it has been implemented in Chapel's copy of Qthreads.

### DIFF
--- a/include/qthread/qthread.hpp
+++ b/include/qthread/qthread.hpp
@@ -14,7 +14,11 @@
 template <bool> class OnlyTrue;
 template <> class OnlyTrue<true>
 {};
+#if __cplusplus >= 201103
+#define QTHREAD_STATIC_ASSERT(X) static_assert((X), #X)
+#else
 #define QTHREAD_STATIC_ASSERT(X) (void)sizeof(OnlyTrue<(bool)(X)>)
+#endif
 #define QTHREAD_CHECKSIZE(X)     QTHREAD_STATIC_ASSERT(sizeof(X) == sizeof(aligned_t))
 
 #define QTHREAD_CHECKINTEGER(X)  QTHREAD_STATIC_ASSERT(std::numeric_limits<X>::is_integer)


include/qthread/qthread.hpp
```diff
@@ -14,7 +14,11 @@
 template <bool> class OnlyTrue;
 template <> class OnlyTrue<true>
 {};
+#if __cplusplus >= 201103
```
m3m0ryh0l3: Is __cplusplus reliable? If not, perhaps change this to:
```
#if defined(__cplusplus) && (__cplusplus >= 201103)
```

dmk42: One nice thing about that test is that if `__cplusplus` is undefined, then the preprocessor has to treat it a zero, therefore it causes the test to do the right thing.

dmk42: BTW, in case you were wondering, that comes from subclause 6.10.1 paragraph 4 of the C standard, and 16.1 paragraph 4 of the C\+\+ standard.  (I'm a standards guy.)

m3m0ryh0l3: Ah, I didn't know that about the preprocessor. I assumed it was either like C, where a non-defined value is a syntax error (I actually think I've run into that problem before, and been forced to wrap things inside an #ifdef)

dmk42: I suspect the case you encountered was a little different, because the behavior is the same in C.  A preprocessor that got that wrong would break a lot of code out there, and wouldn't last long before they had to fix that preprocessor.

npe9:
LGTM

dmk42:
Thanks, guys!  I appreciate it.

npe9:
Happy to help :)

PR #62, opened 2018-09-05
Use Chapel's public memory interface instead of the internal one

ronawho:
We have arranged for Qthreads to get memory from Chapel memory layer
rather than the regular malloc() interface.  But we've been using the
internal memory-layer-specific functions (chpl_malloc() for example)
rather than the public functions (chpl_mem_alloc() for example) that
report out-of-memory conditions and support tracking.  Here, switch
Qthreads' usage to the latter.  This will allow us to get the usual
Chapel "Out of memory allocating <something>" message when the Qthreads
tasking layer asks for memory and the runtime can't supply it.

This modifies the Qthreads src/alloc/chapel.c memory interface to use
the public memory interface instead of the internal one.

Partial cherry-pick of https://github.com/chapel-lang/chapel/commit/0c3b9de163

### DIFF
--- a/src/alloc/chapel.c
+++ b/src/alloc/chapel.c
@@ -15,22 +15,26 @@ static QINLINE int getpagesize()
 
 #include <qthread/qthread-int.h> /* for uint_fast16_t */
 
-#include "chpl-mem-impl.h"
+#include "chpl-mem.h"
+#include "chpl-linefile-support.h"
 
 void *qt_malloc(size_t size){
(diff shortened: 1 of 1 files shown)


ronawho:
Note that the reason we used the internal API originally was because we didn't have a public aligned allocation function. We do now (`chpl_mem_memalign()` added in https://github.com/chapel-lang/chapel/commit/0c3b9de163), so we can target the more public API.

PR #67, opened 2019-02-28
Remove deprecated register keyword

dmk42:
The `register` keyword is deprecated in C++, and some compilers are starting to issue warnings when they see it.  This can cause problems for projects that use `-Werror` because the build aborts upon seeing the warnings.

Chapel is one such project.  In the Chapel runtime, `qthread.h` is included from both C and C++ code, and the C++ code now aborts when compiling for a target such as IA32, for which `qthread.h` contains `register` declarations.

For consistency, this PR removes the `register` keyword everywhere, not just in `qthread.h`.  We'd appreciate it if this PR or something equivalent could be merged.

@gbtitus , @ronawho - FYI.

### DIFF
--- a/include/qt_atomics.h
+++ b/include/qt_atomics.h
@@ -433,7 +433,7 @@ static QINLINE aligned_t qthread_internal_incr_mod_(aligned_t             *opera
     aligned_t retval;
 
 #if QTHREAD_ATOMIC_CAS && (QTHREAD_SIZEOF_ALIGNED_T == 4)
-    register uint32_t oldval, newval;
+    uint32_t oldval, newval;
 
     newval = *operand;
     do {
@@ -444,7 +444,7 @@ static QINLINE aligned_t qthread_internal_incr_mod_(aligned_t             *opera
         newval = __sync_val_compare_and_swap((uint32_t *)operand, oldval, newval);
     } while (oldval != newval);
 #elif QTHREAD_ATOMIC_CAS && (QTHREAD_SIZEOF_ALIGNED_T == 8)
-    register uint64_t oldval, newval;
+    uint64_t oldval, newval;
 
     newval = *operand;
     do {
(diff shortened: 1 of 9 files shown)


m3m0ryh0l3:
`register` is, and remains (as of C11, the most recent iteration of the standard), one of the official 32 C keywords. It should clearly NOT be used in `loop_iter.hpp` (the keyword was deprecated in C++11 and removed in C++17), but it's still useful in C to implement (along with ASM blocks) register variables. Given that this CR touches code across a wide range of architectures, I would be reluctant to remove it without testing the result on those systems.

Does your C++ compiler complain even if the keyword is used inside an `extern "C" {}` block?

dmk42:
Thanks for looking at this.  You are correct that `register` is still a keyword in C.  (The current version is C17; I am the C committee's convener.)

The issue is that sometimes Qthreads headers are included from C++ code, where `register` is deprecated.  In Chapel, this occurs in the runtime.  `extern "C" {}` blocks are processed by the C compiler and not the C++ compiler, so it is not a problem in the generated code, only the runtime code.

dmk42:
It turns out that `register` does not do anything on a declaration that is not an asm statement itself.  That is,
```c
register int *p1 asm ("r0") = ...;
```
does mean something, but
```c
register int *p1;
```
does not do anything even if p1 is later used in an asm statement.

m3m0ryh0l3:
My mistake about C11's recency... though, according to Wikipedia, C17/C18 hasn't (yet?) made any feature changes to C11.

I was under the impression `register` also prevented obtaining the address of the variable (at least, that's what my K&R C book says). And I've always assumed (without evidence) that `register` could also affect the calling convention on systems without a formal ABI (e.g. prevent a function argument from being duplicated onto the stack... which may be handy if we're trying to limit the size of the stack).

dmk42:
Right, C17 is a bugfix version.  (We completed it in 2017 and the __STDC_VERSION__ macro contains 2017, but ISO delayed publication until 2018.)

It's true that `register` does allow the compiler to do some checking to ensure that the variable's address is never taken.  For that feature, it would have been nice to be able to keep it.  I wish C++'s extern "C" meant C syntax and not just C linkage; that's why Clang complains even inside it.

From a functional point of view, `register` does not guarantee anything.  It is just a hint.  I don't know of any compilers less than 25 years old that actually use the hint.  The reason is that people discovered the compiler is better than humans at determining what should live in a register.

npe9:
I'm perfectly okay with removing "register" keywords. As @dmk42 has mentioned they're mostly advisory these days anyway.

olivier-snl:
This seems fine to me.

dmk42:
Thanks!  And I appreciate the questions as well - thanks for being careful before merging.

Issue #70, opened 2020-09-10
Include "local_header" rather than <local_header> in qthread.h


shintaro-iwasaki:
I am Shintaro Iwasaki, an Argonne postdoc who is now working on Open MPI + Qthreads with @olivier-snl. The question is about the way of including headers in `qthread.h`. Now `qthread.h` includes other Qthread headers as follows:
```c
// in qthread/qthread.h
#include <qthread/qthread_int.h>
``` 

However, this causes an issue if a program does not set a path to `qthread.h`. For example,
```c
// in user_prog.c
#include "external/qthread/qthread.h"
```
without `-I QTHERAD_INCLUDE_PATH`, a compiler cannot find `<qthread/qthread_int.h>` since a compiler does not have an include path that contains `qthread/qthread_int.h`.

If `qthread.h` uses a relative path to refer to them, there is no problem.
```c
// in qthread/qthread.h
#include "qthread_int.h"
``` 

For the Open MPI + Qthreads implementation, because of configuration/automake issues in Open MPI, we would like to embed a path to `qthread.h` at configure time without setting `-I` or `CPATH`. I understand that this case is not very usual, but I would like to know the opinions of Qthreads developers about changing a way of including internal headers from `#include <qthread/qthread_xxx.h>` to `#include "qthread_xxx.h"`.

@m3m0ryh0l3: I would truly appreciate it if you could let me know your idea.

I'd be happy if you could tell me a workaround for this if such a change is not allowed. (Needless to say, I will make a PR for this if needed.)

shintaro-iwasaki:
It was addressed by #71. I will close this issue.

PR #71, opened 2020-09-15
Use relative paths to include Qthreads headers in public headers

shintaro-iwasaki:
This PR addresses #70 

Public headers (e.g., `qthread.h`) writes `#include <qthread/xxx.h>` to include other Qthreads headers.  The compiler only looks up global include paths, so a compiler cannot find headers if the Qthreads' include path (`QTHREAD_INSTALL_DIR/include`) is not added to its global include paths.  This patch fixes it by using relative paths (`#include "xxx.h"`).

Note that this change does not break the existing program that includes Qthreads by `#include<qthread.h>` or `#include <qthread/qthread.h>`.

### DIFF
--- a/include/qthread/Makefile.am
+++ b/include/qthread/Makefile.am
@@ -3,7 +3,7 @@
 # Copyright (c)      2008  Sandia Corporation
 #
 
-include_HEADERS = qthread.h
+include_HEADERS = top/qthread.h
 pkginclude_HEADERS = \
 	allpairs.h \
 	barrier.h \
--- a/include/qthread/allpairs.h
+++ b/include/qthread/allpairs.h
@@ -1,8 +1,8 @@
 #ifndef QTHREAD_ALLPAIRS_H
 #define QTHREAD_ALLPAIRS_H
 
-#include <qthread/macros.h>
-#include <qthread/qarray.h>
+#include "macros.h"
+#include "qarray.h"
 
 Q_STARTCXX                             /* */
 typedef void (*dist_f)(const void *unit1, const void *unit2);
--- a/include/qthread/barrier.h
+++ b/include/qthread/barrier.h
@@ -1,8 +1,8 @@
 #ifndef QTHREAD_LOG_BARRIER_H
 #define QTHREAD_LOG_BARRIER_H
 
-#include <qthread/macros.h>
-#include <qthread/qthread.h>
+#include "macros.h"
+#include "qthread.h"
 
 Q_STARTCXX
 /************************************************************/
--- a/include/qthread/cacheline.h
+++ b/include/qthread/cacheline.h
@@ -1,7 +1,7 @@
 #ifndef QTHREAD_CACHELINE_H
 #define QTHREAD_CACHELINE_H
 
-#include <qthread/macros.h>
+#include "macros.h"
 
 Q_STARTCXX                             /* */
 int qthread_cacheline(void);
--- a/include/qthread/dictionary.h
+++ b/include/qthread/dictionary.h
@@ -3,7 +3,7 @@
 #include <inttypes.h>
 #include <stdlib.h>
 
-#include <qthread/macros.h>
+#include "macros.h"
 
 Q_STARTCXX /* */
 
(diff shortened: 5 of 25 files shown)


olivier-snl:
So it looks like you introduced the new "top" directory with a qthread.h to point back to qthread/qthread.h.  I suppose that is the accommodation for Open MPI?  Have you done a "make check" with this change?

shintaro-iwasaki:
Thanks for taking a look at this PR.

I lacked the explanation. It's not for Open MPI. This is needed for this PR.
This top directory is flattened and disappears on installation, so the directory itself is not important.

### The current PR
The current Qthreads "copies" `qthread.h` so `qthread.h` and `qthread/qthread.h` are identical.
```c
// INSTALL_DIR/include/qthread.h
#include <qthread/qthread_int.h>
...

// INSTALL_DIR/include/qthread/qthread.h
#include <qthread/qthread_int.h>
...
```

### This PR without `qthread/top/qthread.h`
This PR rewrites `qthread/qthread.h` (and the other headers in `qthread/`) as follows:
```c
// INSTALL_DIR/include/qthread/qthread.h
#include "qthread_int.h"
...
```
This works if the user includes `#include "INSTALL_DIR/include/qthread/qthread.h"` or `#include <qthread/qthread.h>`. The problem is `#include <qthread.h>`
```c
// INSTALL_DIR/include/qthread.h
#include "qthread_int.h"
...
```
It does not work since the compiler cannot find "qthread_int.h" in `INSTALL_DIR/include/`. To fix it, we need a different header that includes `qthread/qthread.h`

### This PR with `qthread/top/qthread.h`
On installation, this `qthread/top/qthread.h` will be flattened and copied to the "INSTALL_DIR/include" directory since it is added to `include_HEADERS` (https://github.com/Qthreads/qthreads/pull/71/files#diff-59094537dc75d022b32fe91990b899c8), so after installation, it will be like this:
```c
// INSTALL_DIR/include/qthread.h
#include "qthread/qthread.h"

// INSTALL_DIR/include/qthread/qthread.h
#include "qthread_int.h"
...
```
Now the compiler can find "qthread_int.h" even if the user includes `qthread.h`.

The name "top" can be different (e.g., "symbol" or "dummy").

Another possible option is creating a header that has a different name and copying it after renaming that header to "qthread.h" on installation. As far as I quickly checked, `autoconf` does not have such a mechanism (https://www.gnu.org/software/automake/manual/html_node/Headers.html), so I needs to add more code to `Makefile.am`, which might not be clean. I think putting this redirecting `qthread.h` to a directory and flattening it on installation is a cleaner solution.

olivier-snl:
OK, thanks for the explanation.  I think the only potential problem with the change is if Qthreads is compiled into a software package that uses .h files with the same name.  With the global path include, -I<dir> is the first place the compiler looks, but with the local path include, it will first look in the current directory.  Probably not a problem for something like "qloop.h", but seems like it could be a problem for the unfortunately generic-sounding "macros.h"?

shintaro-iwasaki:
The C spec says how to search headers is implementation defined for both <header.h> and "header.h" (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf p150), but GCC and Clang (and I believe most of the other compilers) follow this rule: https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html. If the C compiler follows this rule, there's no problem.

To make the matter more complicated, let me consider the following case.
```
user_prog -+- prog.c
           +- string.h

qthread_inc_dir --- qthread -+- qthread.h
                             +- string.h
another_inc_dir --- string.h
```

`prog.c` includes `qthread.h`, which internally includes `string.h` by `#include "string.h"`. We expect this `string.h` is that of Qthreads.

If I run the following:
```sh
CC prog.c -I$qthread_inc_dir -I$another_inc_dir
# I can reorder -I options.
```
Then, `#include "string.h"` in `qthread.h` can include one of the following:
1. `user_prog/string.h`
2. `qthread_inc_dir/qthread/string.h`
3. `another_inc_dir/string.h`
4. `system's (=Glibc's string.h) string.h`

GCC and Clang choose 2 correctly.

olivier-snl:
Yeah for "implementation defined"... As you know we use it liberally in OpenMP as well.

OK, lets go ahead with the change.

PR #72, opened 2020-09-17
Add FEB non-blocking function declarations to qthread/qthread.h

janciesko:


janciesko:
Todo: remove redundant definitions. This PR becomes necessary if we update the Qthreads interface in Open MPI to properly use FEB semantic. Thanks @olivier-snl for the note.

janciesko:
Obsolete. Can be closed.

PR #74, opened 2020-12-23
Fix compilation errors for hwloc_v2, binders and hwloc_via_chapel when using hwloc2.x

janciesko:
Updates affinity support through hwloc_v2, binders and hwloc_via_chapel to more recent hwloc APIs:
- Fixes compilation errors for hwloc2.x. 
- Maintains backward compatibility to >=hwloc1.5. 
- Uses hwloc2.x fallback where hwloc_get_type_depth(NUMANode) results in a negative int as NUMA nodes are not in the object tree anymore. This is documented [here](https://www.open-mpi.org/projects/hwloc/doc/v2.0.4/a00327.php), section "NUMA level and depth".

### DIFF
--- a/src/affinity/binders.c
+++ b/src/affinity/binders.c
@@ -47,12 +47,12 @@ void INTERNAL qt_affinity_balanced(int num_workers,
     } else {
       // No children
       workers.binds[start] = hwloc_bitmap_alloc();
-      hwloc_bitmap_copy(workers.binds[start], obj->allowed_cpuset);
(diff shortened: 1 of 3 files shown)


olivier-snl (approved):
These changes are mostly allowed_cpuset -> cpuset and HWLOC_OBJ_CACHE to HWLOC_OBJ_CACHE_UNIFIED.  Also a couple of instances of hwloc_set_area_membind_nodeset -> hwloc_set_area_membind. Double checking hwloc.h in the hwloc distribution supports these changes.

PR #84, opened 2021-04-07
Conftest handlesignal qsort

janciesko:
Add signal handling for two conftests in order to avoid segmentation error pop-ups on some platforms when configuring the project.


olivier-snl:
@janciesko : I'll look over this today

olivier-snl (approved):
The code looks good.  I saw that it passed the automated testing.

PR #87, opened 2021-07-01
Support for a custom hwloc symbol prefix

jhh67:
Add a --with-hwloc-symbol-prefix argument to configure that specifiesthe prefix for hwloc
symbols in the case that hwloc was also configured with the --with-hwloc-symbol-prefix argument.
This allows configure to find the hwloc library by searching for a hwloc symbol with the
proper prefix. The default value of this argument is "hwloc_".


janciesko (approved):
LGTM

PR #89, opened 2021-07-19
Syscall fix errno

olivier-snl:
Fixes issue that errno is not preserved following qt system calls.


janciesko:
LGTM. Thanks Stephen.

janciesko (approved):

PR #90, opened 2021-08-04
Mac allow system ucontext

olivier-snl:


ronawho:
@olivier-snl I don't have access to an m1 mac to test on, but I'm wondering if we can automate this detection instead of forcing users to reconfigure.

We have a lot of users coming from a pure python world and many of them are not comfortable or familiar with reconfiguring.

config/qthread_check_swapcontext.m4
```diff
@@ -65,9 +65,9 @@ AC_CACHE_CHECK([whether swapcontext works properly],
   [qthread_cv_swapcontext_works],
   [
   case "$host" in
-  *-darwin*)
-  qthread_cv_swapcontext_works=no
-  ;;
+#  *-darwin*)
```
janciesko, line 68: Since current x86 and M1 Macs seemingly support system ucontext, I think it is ok to drop this case statement.

janciesko: Consider adding something like 
 ```AS_IF([test "$host" in *-darvin*], [qt_host_based_enable_fastcontext=no],
    [qt_host_based_enable_fastcontext=yes]
	;;```
to L37 to disable fast context automatically on M1 Macs while the support is not available yet.

olivier-snl: darvin -> darwin.  But really, it's the combo of darwin and aarch64 (however that parses).  We don't want to disable fast context on earlier macs.

janciesko: Yes, that combo should come together in L37 if I see this correctly.

janciesko: https://github.com/janciesko/qthreads/tree/mac-allow-system-ucontext-2

olivier-snl: > https://github.com/janciesko/qthreads/tree/mac-allow-system-ucontext-2

This is not correct.  m1 macs are arm v8 not arm v7.  The arch to AC will be something like arm64 or aarch64.

janciesko: Yes, this was more an idea / place holder. We'd need to add the right case depending on what config gives us.

olivier-snl: We can match on `aarch64-*-darwin* | arm64*-*-darwin*`

olivier-snl: Actually, since it shouldn't really match any of the architectures in the case, it should fall through without any change.  I just tried from the qthreads repo branch (without the m4 change and without the disable fast context switch on configure) and it worked just fine.

janciesko:
Superseded by #92

PR #92, opened 2021-09-08
Mac allow system ucontext

janciesko:
Removes case block in m4 file. This allows to use ucontext on Macs and removes the requirement to use fast context on all Macs. Using fast context is current unsupported on M1 Macs. This resolves the compilation error.


olivier-snl:
@janciesko The tests are failing because we also need to remove the end of the case statement.

olivier-snl:
These changes look good.  Please check on x86 mac to be sure we didn't mess anything up there -- with no switch to enable or disable fast context it should default to fast context.

I will check on m1 mac.

olivier-snl:
> These changes look good. Please check on x86 mac to be sure we didn't mess anything up there -- with no switch to enable or disable fast context it should default to fast context.
> 
> I will check on m1 mac.

Just tested on m1 and looks good.  Please check on x86 mac to be sure fast context is still used there (and working) by default and also when configured with enable fast context flag set.

janciesko:
Checked and confirming.

PR #93, opened 2021-09-13
Release 1.17 rc

janciesko:
- Remove Loxley schedulers
- Update NEWS
- Add .autogen-version


olivier-snl:
This passes tests but see my comment on the loxley removal about missing distrib in that same list.

PR #95, opened 2021-10-18
Armv8 fastcontext switch

cjhackillinois:
This is the first PR adding ARMv8 custom assembly to implement context switching.


janciesko:
Thanks for the PR. Sincs are reductions and tipically manifest the unrelated runtime bug that we mentioned. Using pthread mutexes instead of the GCC builtins should resolve the problem for the purpose of this PR. You can disable the use of GCC builtins that rely on hw support by configuring with: --disable-hardware-atomics. We might want to wait with benchmarking until I have a fix for the atomics issue ready.

cjhackillinois:
Okay, I can try again with using that command line parameter. To clarify, I run `make check --disable-hardware-atomics -j4`?

olivier-snl:
> Okay, I can try again with using that command line parameter. To clarify, I run `make check --disable-hardware-atomics -j4`?

No, `--disable-hardware-atomics` at configure time

cjhackillinois:
Using this build:`./configure --prefix=/ascldap/users/cdjenk/qt-install --disable-hardware-atomics --with-scheduler=nemesis`

image

Assert does indeed fail as `d_out` and `d_min_authoritative` are different

cjhackillinois:
@janciesko @olivier-snl Just a thought, would saving / restoring the vector registers fix the issue? Floating operations run on the vector unit and ARMv8 does not save / restore this registers.

olivier-snl:
> @janciesko @olivier-snl Just a thought, would saving / restoring the vector registers fix the issue? Floating operations run on the vector unit and ARMv8 does not save / restore this registers.

That is interesting thought.  We could check to see if the compiler is vectorizing.

cjhackillinois:
The qutil.c test uses floating point numbers which mean the usage of the vector / simd unit. See this; Floating-point register organization in AArch64[https://developer.arm.com/documentation/den0024/a/ARMv8-Registers/NEON-and-floating-point-registers/Floating-point-register-organization-in-AArch64]

cjhackillinois:
After reviewing the `qutil_double_min` function, it appears there are some 'sync' calls. We already suspected the syncs may have consistency issues, so not sure if this affects the function above.

janciesko:
Thanks Chris for debugging this. If you run your fast context implementation on ARM as 
`mkdir build && ../configure --with-scheduler=sherwood && make check`, does that succeed repeatedly?

janciesko:
Should fix the issues: https://github.com/Qthreads/qthreads/pull/97

cjhackillinois:
this is what happens:

image

janciesko:
Sounds good. I think we are ready to merge this PR. @olivier-snl

janciesko:
Btw., with PRs: #99 and #98, --with-scheduler={distrib, sherwood} should work now as well. PR #98 should fix the sinc issues that we has seen previously. Feel free to merge those branches into yours for testing.

olivier-snl:
> Btw., with PRs: #99 and #98, --with-scheduler={distrib, sherwood} should work now as well. PR #98 should fix the sinc issues that we has seen previously. Feel free to merge those branches into yours for testing.

Very good work.  Is the order in which we merge these significant?  Any overlapping changes?

janciesko:
auto merge should succeed in any order

janciesko:
This branch sums the current PRs in a testable package. 
QThreads branch: [PRTestBranchNov2021](https://github.com/Qthreads/qthreads/tree/PRTestBranchNov2021)

olivier-snl:
Looking at the arch detection file for the edit @janciesko made for PPC64 reminds me that we seem to still be missing a definition of qthread_gcc_inline_assign for aarch64.  This is just an optimization I believe, not required, but should probably get to it at some point.

cjhackillinois:
@olivier-snl Yes, we spoke about this a month or so ago. We both were not sure how to proceed. If you look inside the file 'config/qthread_check_assembly.m4`, it's missing a aarch 64 definition. It does have an Armv7 definition:

image

olivier-snl:
> @olivier-snl Yes, we spoke about this a month or so ago. We both were not sure how to proceed. If you look inside the file 'config/qthread_check_assembly.m4`, it's missing a aarch 64 definition. It does have an Armv7 definition:
> 
> image

Thanks for reminding me that we spoke about this before.  I don't think it should hold up the PR -- we can go back and look at it later.

cjhackillinois:
I plan to add the single line to the file which should work.

cjhackillinois:
Enabling `qthread_gcc_inline_assign` ended up enabled the pre-processor directive `HAVE_GCC_INLINE_ASSEMBLY`. This lead to other errors since the ARMV8_A64 inline assembly is not defined for certain functions:

image


I believe we have to implement the assembly code.

janciesko:
Thanks Chris, I am aware that we do not have inline asm for some atomics on ARM. I wouldn't worry about this now. We can open an issue if somebody runs into needing this in the future.

src/fastcontext/asm.S
```diff
@@ -782,6 +786,119 @@ SET:
         blr
 #endif
 
+#ifdef NEEDARMA64CONTEXT
+#warning NEEDARMA64CONTEXT
```
janciesko, line 790: Are these warnings still needed?

src/fastcontext/asm.S
```diff
@@ -782,6 +786,119 @@ SET:
...
+#warning GET SET
```
janciesko, line 791: Equally, see comment above.

cjhackillinois: I'm not sure. I originally added it since the Armv7 code had it. I would agree it can be removed.

janciesko: Remove. If output for sanity is required, let's consider a print message at configure.

cjhackillinois: Removed code. I pushed it to the PRDec Branch.

ronawho:
I tested this out with Chapel and it looks good to me. I did correctness and performance testing on a Cray XC with 64-core ThunderX2, an HPE Apollo system with 48-core A64FX, and an AWS cluster with 64-core Graviton2.

Correctness testing looked good on all 3 and I gathered some performance numbers for task spawning and task switching microbenchmarks on them as well a Cray XC with 56-core Skylake as a reference. The task creation microbenchmark just times creating and waiting for completion of numCores tasks 500K times. The task switching microbenchmark creates numCores * 4 tasks and just yields between them 500K times:


```chpl
use Time;

config const trials = 500000;

proc main() {
  var t: Timer; t.start();

  for 1..trials do
    coforall 1..here.maxTaskPar { }
  writef("%i task creation: %.2dr\n", trials*here.maxTaskPar, t.elapsed()); t.clear();

  var total: atomic int;
  coforall 1..here.maxTaskPar*4 {
    var i: int;
    for 1..trials {
      i += 1;
      chpl_task_yield();
    }
    total.add(i);
  }
  writef("%i task switching: %.2dr\n", total.read(), t.elapsed()); t.clear();
}
```

Here's the performance results:

Task Creation:

| Config            | No ASM | ASM    |
| ----------------- | -----: | -----: |
| 64-core Skylake   | 12.95s | 11.23s |
| 64-core ThunderX2 | 24.49s | 22.86s |
| 48-core A64FX     | 25.08s | 24.00s |
| 64-core Graviton2 | 10.01s |  9.34s |

Task Switching:

| Config            | No ASM | ASM    |
| ----------------- | -----: | -----: |
| 64-core Skylake   |  4.30s | 0.24s  |
| 64-core ThunderX2 |  4.85s | 0.36s  |
| 48-core A64FX     |  3.70s | 1.12s  |
| 64-core Graviton2 |  1.65s | 0.28s  |

So minor improvements for task creation (not surprising, more goes into creating and scheduling tasks than just the makecontext) and generally speaking large improvements for task switching. 

I didn't run a ton of real benchmarks, but I did check a few that do a lot of fine-grained communication (where we task switch while waiting for comm to complete) and I saw a ~5% improvement for them (~78MB/s per node to ~82MB/s/node for ordered bale indexgather.)

We saw nice improvements for a variety of benchmarks in the past when optimizing task creation (https://chapel-lang.org/releaseNotes/1.15/09-perf-opts.pdf) so I expect the minor task creation improvements in this PR to benefit those same benchmarks.

src/fastcontext/asm.S
```diff
@@ -782,6 +786,117 @@ SET:
...
+        ldp     Q28, Q29, [X0,#448]
+        ldp     Q30, Q31, [X0,#480]
+        sub     X0, X0, #256
+#endif
+        ldr     X0, [X0]
+	ret
+#endif
+
 #ifdef NEEDARMCONTEXT
 #warning NEEDARMCONTEXT
```
janciesko, line 901: Let's remove these two warnings as well

janciesko:
Thanks for looking into this @ronawho. "No ASM" vs "ASM" corresponds to Qthreads builds with and without "--disable-fastcontext"?

janciesko:
Thanks @ronawho. https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/x86_64/swapcontext.S.html#251 shows indeed a syscall in line 80.

janciesko:
LIBC for ARM shows this implementation for swapcontext: https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/arm/swapcontext.S.html. For my understanding - @cjhackillinois, do you see a system call in this ASM code? Reason is, I am trying to understand the perf diff of the above Chapel benchmark runs on ARM.

janciesko:
@cjhackillinois Idea: would it make sense to create a `ncontext`* GitHub project which could be used in projects such as LLVM, Argobots and qthreads? Would this be related to https://swtch.com/libtask/ ?
\* nanocontext just to follow the naming idea

dmk42:
> https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/x86_64/swapcontext.S.html#251 and https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/aarch64/swapcontext.S.html indeed show syscalls.

Oh, good point.  POSIX 2.4.1 paragraph 5 says "Each thread has a 'signal mask' that defines the set of signals currently blocked from delivery to it."

https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/V2_chap02.html#tag_15_04_01

If Qthreads does not save and restore the signal mask upon context switch, then it must forbid each qthread from calling `sigaction()`.  I don't think that's the intention, so I think that means the fast context switches, in their current form, should someday lead to a race condition.  That would mean they are currently faster because they are incorrect.

Did I misunderstand?

dmk42:
I did misunderstand, but unfortunately, not in a way that helps.  `sigaction()` affects the entire process, so forbidding it would not help.

When a process receives a signal, it is delivered to exactly one thread.  In the normal case, the signal becomes blocked for that thread while it is being handled, thus the need for per-thread signal masks.  There is a possibility of a thread being interrupted and context-switched while it is in a signal handler, so the per-thread mask needs to be saved and restored.  That is what the syscalls do in the library version of `getcontext()`/`setcontext()`.

There are some architectural decisions that could be made to remove the need to save and restore the thread's signal mask, including forbidding `pthread_sigmask()`, blocking all signals to context-switched qthreads so that signals are only delivered to threads that will not be context switched by the Qthreads package, and possibly other things.  I doubt if that is what you would want.

Consequently, there will not be much if any performance advantage to supplying hand-written context switch assembly language, since to be correct, it must make the same syscalls that the library routines make.

That doesn't mean there is no need for Qthreads to supply context switching code.  Android does not supply `getcontext()`/`setcontext()`, and macOS deprecates them.  There are probably other "almost POSIX" systems where they are not available.

Given that Qthreads-supplied context switching is useful for those systems, it might as well be the default everywhere, so that it will be heavily exercised to make sure there are no bugs.

janciesko:
Thanks David for the comment and insight. Note that in user-level threading, a context swap is a user-level operation. It is oblivious to system interrupts. Ucontext is a fall-back option in case we do not provide a user-level implementation of that functionality on a given architecture.

ronawho:
I don't follow a lot of the conversation about signals, but for 

> Consequently, there will not be much if any performance advantage to supplying hand-written context switch assembly language, since to be correct, it must make the same syscalls that the library routines make.

I can't imagine that's right. There are a lot of libraries (e.g. argobots) and languages (e.g. go) that do task switching in user-space to avoid kernel overheads and so far as I know they end up using native ASM. I cannot imagine that some syscalls are needed for correct implementation.

dmk42:
Signals are a user-level operation, not something that is limited to the kernel.  In general code, the user level still has to take the signal mask into account.

Go has the advantage of defining what is allowed and what is not allowed in a Goroutine, so they have the ability to get around the problem by restricting the code that is being context switched.  In contrast, if I understand correctly, Qthreads typically context switches general C or C++ code, which can often use signals.

Argobots does indeed do syscalls during their context switches when they are running general, unrestricted code.  Only when the code is restricted can they dispense with the syscalls.  In other words, they provide two different semantics, and can optimize away the syscalls in the cases when the user chooses the more restrictive semantics.  This is explained in the following paper.

https://www.mcs.anl.gov/~aamer/papers/tpds17_argobots.pdf

From section 2.4 of the paper:  "Our implementation exploits fcontext by default and provides ucontext as an alternative when the user requires preserving the signal mask between context switches. Indeed, fcontext is significantly faster than ucontext mostly because it avoids preserving the signal mask, which requires expensive system calls."

Qthreads needs to address signals in some way, either by preserving the signal mask during context switches, or by restricting the code that can be executed, like Argobots does.

It might be helpful to involve someone who does a lot of work with signals, such as @gbtitus .

ronawho:
Ah, thanks for the additional info and the argobots quote @dmk42 -- that's really helpful.

Unrelated, but FYI that I just got access to an M1 based mac. Are we expecting this assembly to work for that too?

If so I would guess `src/fastcontext/asm.S` needs to be updated. Trying that naively I get errors about `.type` being an unknown directive. Commenting that out I can build, but hellos seem to be hanging.

olivier-snl:
> Ah, thanks for the additional info and the argobots quote @dmk42 -- that's really helpful.
> 
> Unrelated, but FYI that I just got access to an M1 based mac. Are we expecting this assembly to work for that too?
> 
> If so I would guess `src/fastcontext/asm.S` needs to be updated. Trying that naively I get errors about `.type` being an unknown directive. Commenting that out I can build, but hellos seem to be hanging.

Apple "macho" and Linux "elf" assembly are slightly different.  We've been concentrating on getting things squared away for arm64 Linux first.

ronawho:
re the signal handling -- I attached the qthreads paper I've collected over the years (probably available online too, but I didn't feel like re-finding non-paywall links):

- [LIGHTWEIGHT THREADING FOR ARCHITECTURAL DESIGN RESEARCH.pdf](https://github.com/Qthreads/qthreads/files/7893019/LIGHTWEIGHT.THREADING.FOR.ARCHITECTURAL.DESIGN.RESEARCH.pdf)
  > "The API’s goal is to add as little overhead as possible to experimental architectural threading primitives while maintaining its generic nature. For example, to reduce overhead, qthreads are considered mostly anonymous, and cannot be affected directly by other threads—there is no way to cancel a currently running thread, and threads are not expected to respond to signals"
- [Scheduling Task Parallelism on Multi-Socket Multicore Systems .pdf](https://github.com/Qthreads/qthreads/files/7893022/Scheduling.Task.Parallelism.on.Multi-Socket.Multicore.Systems.pdf)
  >  "The Qthreads lightweight threading concept is intended to match future hardware threading environments more closely than existing concepts in three crucial aspects: anonymity, introspectable limited resources, and inherent localization. Unlike heavyweight threads, these threads do not support expensive features like per-thread identifiers, per thread signal vectors, or preemptive multitasking."
 - [Qthreads- An API for Programming with Millions of Lightweight Threads .pdf](https://github.com/Qthreads/qthreads/files/7893020/Qthreads-.An.API.for.Programming.with.Millions.of.Lightweight.Threads.pdf)

janciesko:
Good insights. Makes sense. We have this config option description - see below. Here "If we you run into bugs" covers running the unrestricted code :) We could update the description. 
```
--disable-fastcontext     use a lighter-weight non-system-based context
                          swapping mechanism that does not make system calls.
                          If you run into bugs, you can disable it on some
                          systems to use the slower libc-provided version.
```

dmk42:
@ronawho - Thanks.  That totally addresses the issue, by restricting what is allowed in context-switched code.

> Unrelated, but FYI that I just got access to an M1 based mac. Are we expecting this assembly to work for that too?

+1
I started watching this for exactly the same reason.  The stepwise refinement makes sense, getting it stable on Linux before addressing the Mac.

dmk42:
> Good insights. Makes sense. We have this config option description - see below. Here "If you run into bugs" covers running the unrestricted code :) We could update the description. Should we?
> 
> ```
> --disable-fastcontext     use a lighter-weight non-system-based context
>                           swapping mechanism that does not make system calls.
>                           If you run into bugs, you can disable it on some
>                           systems to use the slower libc-provided version.
> ```

Since Qthreads addresses the issue by design, the existing documentation is technically sufficient, but adding to the documentation would prevent people from doing what I did and thinking there is a problem.

ronawho:
> Since Qthreads addresses the issue by design, the existing documentation is technically sufficient, but adding to the documentation would prevent people from doing what I did and thinking there is a problem.

@dmk42 could you say what you were trying to do, or if you have a code example that would be great. Chapel doesn't have much signal handling support today, but https://github.com/chapel-lang/chapel/issues/13645 is a longstanding feature request and I want to make sure I understand if this limits any future design decisions we make (and probably obvious, but I have very little experience with signals.)

I know that in the runtime we have a few places with signal handling, but maybe straightforward signal handling is ok? And the issue is just that signals aren't preserved on task switching or maybe instead that signals are held on the thread boundary so switching tasks results in the new tasks having the same signals?

dmk42:
Sorry, I didn't mean that I had code that failed.  What I did was just read the Qthreads code and notice that it didn't preserve the signal mask across context switches.

When a process receives a signal, a thread with that signal unblocked is selected essentially at random to receive the signal.  Therefore, to implement https://github.com/chapel-lang/chapel/issues/13645 , one of two things needs to happen.  Either the slower context switch needs to be used so that the signal mask is preserved, or all the threads that are potentially context-switched by Qthreads need to block all signals, so that some other thread will receive any incoming signal. 
 In the latter case, Chapel would need to provide some way to ensure that there is one thread somewhere that is not managed by Qthreads so that the signal handler can run on it.

Here is a scenario that could trip up the proposed Chapel feature, if the slower context switching is not used.

A Qthreads-managed thread installs a signal handler and unblocks the associated signal.  Some context switches happen, and the unblock floats around among the qthreads.  Later in the program, the original thread is finished allowing that signal, and it blocks the signal.  However, the original unblock has now floated to some other thread, so the blocking operation is a no-op.  The signal cannot be blocked because nobody knows which thread has it unblocked at this point.

dmk42:
> I know that in the runtime we have a few places with signal handling, but maybe straightforward signal handling is ok? And the issue is just that signals aren't preserved on task switching or maybe instead that signals are held on the thread boundary so switching tasks results in the new tasks having the same signals?

A cursory search through Chapel shows that `sighold()` and `sigrelse()` are not used.  I believe the only two other functions that block and unblock signals are `sigprocmask()` and `pthread_sigmask()`.  (I hope that's right.)

`sigprocmask()` is only defined for single-threaded programs.  It does not appear in Chapel itself, but does appear in some third-party libraries that it uses, so it might be worth checking whether it is ever called by those packages from within Chapel.

There are some places in the Chapel runtime and third-party libraries that use `pthread_sigmask()` to block and unblock signals.  If those calls happen on threads that are not managed by Qthreads, then there is definitely no problem.  Otherwise, the problem scenario I outlined above could be an issue.  It's worth checking.

dmk42:
[This is a cleanup of two prior postings due to new information.  I'll delete the old ones.]

Sorry, I didn't mean that I had code that failed. What I did was just read the Qthreads code and notice that it didn't preserve the signal mask across context switches.

Based on Brad's model in https://github.com/chapel-lang/chapel/issues/13645#issuecomment-669669180 , I believe the proposed user-level Chapel feature will not have problems.  That leaves the runtime, so I'll lay out some conditions that might cause the Chapel runtime to have problems.

> I know that in the runtime we have a few places with signal handling, but maybe straightforward signal handling is ok? And the issue is just that signals aren't preserved on task switching or maybe instead that signals are held on the thread boundary so switching tasks results in the new tasks having the same signals?

When a process receives a signal from the outside (not due to a `raise()`), a thread with that signal unblocked is selected essentially at random to receive the signal.  There are a couple of ways to deal with this.  The slower context switch can be used so that the signal mask is preserved, or all the threads that are potentially context-switched by Qthreads can block all signals, so that some other thread can receive any incoming signal.  In the latter case, Chapel could provide some way to ensure that there is one thread somewhere that is not managed by Qthreads so that the signal handler can run on it.

Here is a scenario that could run into trouble with the fast context switch.

A Qthreads-managed thread installs a signal handler and unblocks the associated signal. Some context switches happen, and the unblock floats around among the qthreads. Later in the program, the original thread is finished allowing that signal, and it blocks the signal. However, the original unblock has now floated to some other thread, so the blocking operation is a no-op. The signal cannot be blocked because nobody knows which thread has it unblocked at this point.

A cursory search through Chapel shows that `sighold()` and `sigrelse()` are not used. I believe the only two other functions that block and unblock signals are `sigprocmask()` and `pthread_sigmask()`. (I hope that's all.)

`sigprocmask()` is only defined for single-threaded programs. It does not appear in Chapel itself, but does appear in some third-party libraries that it uses, so it might be worth checking whether it is ever called by those packages from within Chapel.

There are some places in the Chapel runtime and third-party libraries that use `pthread_sigmask()` to block and unblock signals. If those calls happen on threads that are not managed by Qthreads, then there is definitely no problem. Otherwise, the problem scenario I outlined above could be an issue. It's worth checking.

olivier-snl:
> @ronawho - Thanks. That totally addresses the issue, by restricting what is allowed in context-switched code.
> 
> > Unrelated, but FYI that I just got access to an M1 based mac. Are we expecting this assembly to work for that too?
> 
> +1 I started watching this for exactly the same reason. The stepwise refinement makes sense, getting it stable on Linux before addressing the Mac.

Yes, and in the mean time, I have seen our initial tests pass on m1 mac using CFLAGS=-D_XOPEN_SOURCE and qthreads configure options --disable-fastcontext --with-default-stack-size=65536.  That will use the (deprecated) system-provided ucontext, and requires a larger stack size than we usually use.  This configuration has not been robustly tested though, and we do intend to proceed on the asm-based context switching.  The cursory testing was on my wife's personal m1 mac, because we did not have access to one at work.

gbtitus:
> When a process receives a signal from the outside (not due to a `raise()`), a thread with that signal unblocked is selected essentially at random to receive the signal.

This is already implied by what @dmk42 wrote but for completeness I'll just add that "due to a `raise()`" is really "due to something the thread itself did." I.e., if a thread does `raise(SIGUSR1)` the signal is delivered specifically to itself, but also if a thread does `return *((int*) 0);` (thus dereferencing typically-nonexistent memory), the corresponding `SIGSEGV` is delivered specifically to that same thread.

It's kind of like worst-case classroom discipline: if the teacher can tell which student misbehaved they punish that one, but if they know misbehavior occurred but not by whom, they just punish a randomly-selected student. (Extending this analogy to include the effects of the signal masks is left as entertainment for the reader.)

David, I would have contributed earlier, but I'm down to 50% time pending retirement and we're trying to treat me as a back-room resource/consultant.  😀

dmk42:
An offline discussion has revealed that Chapel does not allow Qthreads to migrate a worker thread once it is running.  I believe that solves the problem, because then the worker thread remains bound to the pthread with the expected signal mask.

That means for Chapel, there is no semantic difference between the fast and slow context switch.

src/fastcontext/asm.S
cjhackillinois: @janciesko I removed the ARMv7 warnings above. Commit: 09525efd6c9820d84e26a9200066baff59c919eb

cjhackillinois:
Just for completeness:
Does slow context switch = ucontext library and fast context switch = arch-specific native assembly?

Given that qthreads is a ULT computing model, I believe any application which requires signal support should just use pthreads.  Qthreads are better suited for applications which don't require signals IMO. This keeps the qthread state small.

dmk42:
> Just for completeness:
> Does slow context switch = ucontext library and fast context switch = arch-specific native assembly?

Right, that's what I meant.



> Given that qthreads is a ULT computing model, I believe any application which requires signal support should just use pthreads. Qthreads are better suited for applications which don't require signals IMO. This keeps the qthread state small.

In general, I think this is right, though Chapel requires signals, and Chapel has restricted its use of Qthreads such that it can get away with not saving the signal mask (because in Chapel, each qthread remains bound to the same pthread).

cjhackillinois:
Makes sense.

cjhackillinois:
@olivier-snl Any concerns with me deleting this branch?

PR #97, opened 2021-11-17
Change calls to QTHREAD_TRYLOCK_{INIT,DESTROY} to pass-by-ref

janciesko:
QTHREAD_TRYLOCK_INIT and QTHREAD_TRYLOCK_DESTROY initialized and destroyed local stack variables. This seems like unwanted behavior. Occasionally and interestingly on x86, uninitialized locks happened to be zero-initialized which resulted in working code.


olivier-snl:
Good work tracking this down.

cjhackillinois:
why does passing by ref fix the issue?

Issue #57, opened 2018-07-05
sinc issues with multiple workers per shepherd on power and arm


stelleg:
I'm seeing issues with `sherwood` and `distrib` on power and arm when using sincs and `NUM_WORKERS_PER_SHEPHERD > 1`.

I'm guessing there needs to be some added memory fences. A sinc-based fib exercises the bugs:

```
#include <qthread/qthread.h>
#include <qthread/sinc.h>
#include <assert.h>

typedef struct {
  int i;
  qt_sinc_t* s;
  int *ret;
} args_t;

aligned_t fib(void *args_){
  args_t* args = (args_t*)args_;
  if(args->i < 2){
    *args->ret = 1;
    qt_sinc_submit(args->s, NULL);
    return 0;
  }
  int x, y;
  qt_sinc_t sinc;
  qt_sinc_init(&sinc, 0, NULL, NULL, 2);

  args_t argsx = { args->i-1, &sinc, &x };
  args_t argsy = { args->i-2, &sinc, &y };

  qthread_fork_copyargs(fib, &argsx, sizeof(args_t), NULL);
  qthread_fork_copyargs(fib, &argsy, sizeof(args_t), NULL);

  qt_sinc_wait(&sinc, NULL);
  qt_sinc_submit(args->s, NULL);
  *args->ret  = x + y;
  return 0;
}

int main(){
  assert(qthread_initialize() == 0);
  qt_sinc_t sinc;
  qt_sinc_init(&sinc, 0, NULL, NULL, 1);
  int ret;
  args_t args = { 30, &sinc, &ret };
  qthread_fork_copyargs(fib, &args, sizeof(args_t), NULL);
  qt_sinc_wait(&sinc, NULL);
  printf("%d\n", ret);
}
```

stelleg:
Looks like `distrib` can also have  issues even with 1 worker per shepherd. I'm also not convinced it is `sinc`-related, as this errors seem to be occurring in enqueue dequeue subroutines.

npe9:
Gotcha, I'll look at this next week.

olivier-snl:
@stelleg Thanks for the reproducer, which we might want to add to the test suite.  The timing is good since we are preparing for an end of FY point release, and we can include a fix for this in that.

npe9:
@olivier-snl I just added @stelleg's test to the test suite (it passes on OSX). I'll play with it on the testbeds. Do we have a testbed that would be particularly representative?

[...]

PR #98, opened 2021-11-18
Fix memory consistency issues on PowerPC and ARM

janciesko:
Adds the atomic type specifier (C11) to members of structs used for thread synchronization. This implements sequential consistency (memory_order_seq_cst) for loads and stores to these members. This PR should fix issues on weakly-ordered systems (ARM, PowerPC) as described here:
- Sinc issues with multiple workers per shepherd on power and arm [#57]

### DIFF
--- a/include/qt_atomics.h
+++ b/include/qt_atomics.h
@@ -5,6 +5,7 @@
 
 #include <qthread/common.h>
 #include <qthread/qthread.h>
+#include <stdatomic.h>
 
 #ifdef QTHREAD_NEEDS_IA64INTRIN
 # ifdef HAVE_IA64INTRIN_H
@@ -26,8 +27,8 @@
 #elif (QTHREAD_ASSEMBLY_ARCH == QTHREAD_IA32) || \
     (QTHREAD_ASSEMBLY_ARCH == QTHREAD_AMD64)
 # define SPINLOCK_BODY() do { COMPILER_FENCE; __asm__ __volatile__ ("pause" ::: "memory"); } while (0)
-#else // ifdef QTHREAD_OVERSUBSCRIPTION
-# define SPINLOCK_BODY() do { COMPILER_FENCE; } while (0)
+#else 
+# define SPINLOCK_BODY() do { COMPILER_FENCE; atomic_thread_fence(memory_order_acquire); /*asm volatile("sync");*/ } while (0)
 #endif // ifdef QTHREAD_OVERSUBSCRIPTION
 
 #if defined(__tile__)
@@ -56,7 +57,9 @@ void qt_spin_exclusive_unlock(qt_spin_exclusive_t *);
 # define QTHREAD_FASTLOCK_LOCK(x)     { aligned_t val = qthread_incr(&(x)->enter, 1); \
                                         while (val != (x)->exit) SPINLOCK_BODY(); /* spin waiting for my turn */ }
 # define QTHREAD_FASTLOCK_UNLOCK(x)   do { COMPILER_FENCE; \
-                                           (x)->exit++; /* allow next guy's turn */ } while (0)
+                                           (x)->exit++; /* allow next guy's turn */ \
+                                           atomic_thread_fence(memory_order_release); \
+                                        } while (0)
 # define QTHREAD_FASTLOCK_DESTROY(x)
 # define QTHREAD_FASTLOCK_DESTROY_PTR(x)
 # define QTHREAD_FASTLOCK_TYPE        qt_spin_exclusive_t
@@ -132,9 +135,10 @@ typedef union qt_spin_trylock_s {
 # define QTHREAD_TRYLOCK_INIT_PTR(x) { (x)->u = 0; }
 # define QTHREAD_TRYLOCK_LOCK(x)     { uint32_t val = qthread_incr(&(x)->s.users, 1); \
                                        while (val != (x)->s.ticket) SPINLOCK_BODY(); /* spin waiting for my turn */ }
-# define QTHREAD_TRYLOCK_UNLOCK(x)   do { COMPILER_FENCE;                                               \
+# define QTHREAD_TRYLOCK_UNLOCK(x)   do { COMPILER_FENCE; \
                                           qthread_incr(&(x)->s.ticket, 1); /* allow next guy's turn */ \
-} while (0)
+                                          atomic_thread_fence(memory_order_release); \
+                                        } while (0)
 # define QTHREAD_TRYLOCK_DESTROY(x)
 # define QTHREAD_TRYLOCK_DESTROY_PTR(x)
 
--- a/include/qt_atomics.h
+++ b/include/qt_atomics.h
@@ -5,7 +5,19 @@
 
 #include <qthread/common.h>
 #include <qthread/qthread.h>
+
+#ifndef __STDC_NO_ATOMICS__
+#define USE_C11_MEMORY_FENCE
 #include <stdatomic.h>
+#endif
+
+#ifdef USE_C11_MEMORY_FENCE
+#define THREAD_FENCE_MEM_ACQUIRE atomic_thread_fence(memory_order_acquire)
+#define THREAD_FENCE_MEM_RELEASE atomic_thread_fence(memory_order_release)
+#else
+#define THREAD_FENCE_MEM_ACQUIRE MACHINE_FENCE
+#define THREAD_FENCE_MEM_RELEASE MACHINE_FENCE
+#endif
 
 #ifdef QTHREAD_NEEDS_IA64INTRIN
 # ifdef HAVE_IA64INTRIN_H
@@ -56,10 +68,10 @@ void qt_spin_exclusive_unlock(qt_spin_exclusive_t *);
 # define QTHREAD_FASTLOCK_INIT_PTR(x) { (x)->enter = 0; (x)->exit = 0; }
 # define QTHREAD_FASTLOCK_LOCK(x)     { aligned_t val = qthread_incr(&(x)->enter, 1); \
                                         while (val != (x)->exit) SPINLOCK_BODY(); \
-                                          atomic_thread_fence(memory_order_acquire); /* spin waiting for my turn */ }
+                                          THREAD_FENCE_MEM_ACQUIRE; /* spin waiting for my turn */ }
 # define QTHREAD_FASTLOCK_UNLOCK(x)   do { COMPILER_FENCE; \
                                            (x)->exit++; /* allow next guy's turn */ \
-                                           atomic_thread_fence(memory_order_release); \
+                                           THREAD_FENCE_MEM_RELEASE; \
                                         } while (0)
 # define QTHREAD_FASTLOCK_DESTROY(x)
 # define QTHREAD_FASTLOCK_DESTROY_PTR(x)
@@ -136,10 +148,10 @@ typedef union qt_spin_trylock_s {
 # define QTHREAD_TRYLOCK_INIT_PTR(x) { (x)->u = 0; }
 # define QTHREAD_TRYLOCK_LOCK(x)     { uint32_t val = qthread_incr(&(x)->s.users, 1); \
                                        while (val != (x)->s.ticket) SPINLOCK_BODY(); \
-                                          atomic_thread_fence(memory_order_acquire); /* spin waiting for my turn */ }
+                                          THREAD_FENCE_MEM_ACQUIRE; /* spin waiting for my turn */ }
 # define QTHREAD_TRYLOCK_UNLOCK(x)   do { COMPILER_FENCE; \
                                           qthread_incr(&(x)->s.ticket, 1); /* allow next guy's turn */ \
-                                          atomic_thread_fence(memory_order_release); \
+                                          THREAD_FENCE_MEM_RELEASE; \
                                         } while (0)
 # define QTHREAD_TRYLOCK_DESTROY(x)
 # define QTHREAD_TRYLOCK_DESTROY_PTR(x)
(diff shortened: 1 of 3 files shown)


olivier-snl:
Chapel currently recommends but does not require C11, FYI.

dmk42:
Certain configurations of Chapel do require C11, so I would recommend checking with @ronawho .  They might be willing to require C11 all the time, to get the improved ARM support.

stelleg:
Just tested this quickly on a 64 core ThunderX2 system and am seeing the following failures:

```
FAIL lock_acq_rel (exit status: 139)
FAIL task_spawn (exit status: 139)
FAIL test_spawn_simple (exit status: 139)
```

janciesko:
Yes, I can reproduce. Calls too `atomic_thread_fence(memory_order_seq_cst);` seem to have no effect. However, making the lock variables _Atomic,  see permalink below, seems to fix the issue but would fail at compilation on MacOS/CLANG (Atomic built-ins require PODs or pointer to pointer).
https://github.com/Qthreads/qthreads/blob/71d29efea190c7cd1349be9b4feb0d5047f92132/include/qt_atomics.h#L125-L126

janciesko:
An issue reproducer is located here: https://github.com/janciesko/qthreads/commit/8900fc3734bb3da6c56df46d5efa8c5e98aa03a9

It shows consistent failures on ARM and Power ISA and succeeds on both architectures if USE_PTHREAD is enabled. It succeeds consistently in either case on x86.

janciesko:
I have updated the PR by moving the atomic_thread_fence(memory_order_acquire) to the outside of the spin lock body. This was an error. The current PR version fixes the issue in my test runs.

janciesko:
@olivier-snl @cjhackillinois
Qthreads builds and passes on my x86 Mac (Clang 13) with this PR. Further, the following configurations pass on all 
test beds;
`--disable-builtin-atomics //use asm assembly`
`--disable-hardware-atomics //use fallback via pthread mutex`
Note: The inline assembly code includes `isync` on the PowerPC ISA which removes the consistency issue. We do not have asm implementations of locks for ARM thus either pthread mutexes or built-in atomics must be used.

cjhackillinois:
@janciesko To confirm, PR 98 and 99 add fixes so builtins on ARMv8 work?

olivier-snl:
> @janciesko To confirm, PR 98 and 99 add fixes so builtins on ARMv8 work?

Yes, along with your #95

include/qt_atomics.h
```diff
@@ -6,6 +6,19 @@
 #include <qthread/common.h>
 #include <qthread/qthread.h>
 
+#ifndef __STDC_NO_ATOMICS__
+#define USE_C11_MEMORY_FENCE
+#include <stdatomic.h>
+#endif
```
ronawho, line 12: I'm not sure how much this helps with portability. There's only a few compilers that actually set `__STDC_NO_ATOMICS__` (I think intel for a release or 2 and PGI for a while) and obviously compilers that were out before C11 was introduced won't have this set.

I don't think this is an issue for Chapel, as we now require fairly modern C compilers, but it may hurt overall qthreads portability. Maybe this could just be a configure test instead?

ronawho:
Hi @dmk42 -- good to hear from you, been a while!

---

I haven't thought about this issue in some time, so catching myself back up:
 - We saw hangs and failures for several Chapel tests on an internal ARM XC system ever since we started testing on it. I think we worked around this in April 2020 by not using native FEB in https://github.com/chapel-lang/chapel/pull/15520. Reverting that PR, I still see failures for our test `$CHPL_HOME/test/runtime/configMatters/comm/cache-remote/acqrel-sync.chpl` (fails ~1/25 trials)
 - I don't remember how I got to the reproducer, but I think I sent Jan the test in this PR back in April 2021 as something that failed for me on our internal machine and Mayer (Sandia ARM machine.) I haven't verified, but I'm assuming that test now passes.

---

Testing now, `$CHPL_HOME/test/runtime/configMatters/comm/cache-remote/acqrel-sync.chpl` seems to pass for me with this PR applied and https://github.com/chapel-lang/chapel/pull/15520 reverted (at least for 100 trials, I'll run more overnight.) That said, I do worry that there could still be other memory consistency issues lurking. Have you done an audit of the code, or mostly just focusing on the known test failures?

I also worry this is adding an extra fence on x86 where it might not be needed. I've set up testing in Chapel's performance playground and will report back results in the next day or 2 after that gets a chance to run. If there are performance regressions I wonder if you could reimplement the ticket counter purely with C11 atomics (likely with orders beyond `seq_cst` to allow the backend to best optimize for each ISA.)

include/qt_atomics.h
janciesko: We could add a configure check in favor of less ifdefs in the code. The effect would be the same. Clang and MSVC do set this variable.

janciesko:
Thanks Elliott. I have audited the code. Have you observed any indication for another memory inconstancy?
The implementation requires a thread_fence semantic on entering and exiting the critical section in order to propagate memory updates that occurred within the section. This comes for free on x86 as a side effect of atomic incr of the ticket counter. We could thus omit the thread barrier on x86 if we see a performance regression.

ronawho:
Unfortunately with more trials I'm still seeing errors for our test. With 1000 trials, I saw 3 wrong answers and 2 timeouts (presumably hangs) for `$CHPL_HOME/test/runtime/configMatters/comm/cache-remote/acqrel-sync.chpl`

If it's helpful I can try to get instructions for how to reproduce on Mayer, but I don't think I'll have time to help create a standalone qthreads reproducer until after the new year.

include/qt_atomics.h
olivier-snl: > 

To support earlier compilers, we could guard on __STDC_VERSION__ (if you don't want to work with the configure test)

ronawho:
I wanted to see if the issue for my test is with these locks or something other code. I changed to a simple C11 test-and-test-and-set lock (taken from a spinlock we use in the chapel runtime, it's fast but doesn't have the fairness properties that the ticket lock does) and with that my test passed 1000 trials. To me that indicates there's still some memory consistency issue in this updated lock implementation.

Diff on top of this PR:

```diff
diff --git a/qthread-src/include/qt_atomics.h b/qthread-src/include/qt_atomics.h
index 322b8cc56f..6132e98abe 100644
--- a/qthread-src/include/qt_atomics.h
+++ b/qthread-src/include/qt_atomics.h
@@ -9,6 +9,7 @@
 #ifndef __STDC_NO_ATOMICS__
 #define USE_C11_MEMORY_FENCE
 #include <stdatomic.h>
+#include <stdbool.h>
 #endif

 #ifdef USE_C11_MEMORY_FENCE
@@ -58,25 +59,17 @@
 #elif defined(USE_INTERNAL_SPINLOCK) && USE_INTERNAL_SPINLOCK
 # define QTHREAD_FASTLOCK_SETUP() do { } while (0)
 # define QTHREAD_FASTLOCK_ATTRVAR
-typedef struct qt_spin_exclusive_s { /* added to allow fast critical section ordering */
-    aligned_t enter;                 /* and not call pthreads spin_lock -- hard to debug */
-    aligned_t exit;                  /* near the lock under gdb -- 4/1/11 akp */
-} qt_spin_exclusive_t;
+typedef atomic_bool qt_spin_exclusive_t;
 void qt_spin_exclusive_lock(qt_spin_exclusive_t *);
 void qt_spin_exclusive_unlock(qt_spin_exclusive_t *);
-# define QTHREAD_FASTLOCK_INIT(x)     { (x).enter = 0; (x).exit = 0; }
-# define QTHREAD_FASTLOCK_INIT_PTR(x) { (x)->enter = 0; (x)->exit = 0; }
-# define QTHREAD_FASTLOCK_LOCK(x)     { aligned_t val = qthread_incr(&(x)->enter, 1); \
-                                        while (val != (x)->exit) SPINLOCK_BODY(); \
-                                          THREAD_FENCE_MEM_ACQUIRE; /* spin waiting for my turn */ }
-# define QTHREAD_FASTLOCK_UNLOCK(x)   do { COMPILER_FENCE; \
-                                           (x)->exit++; /* allow next guy's turn */ \
-                                           THREAD_FENCE_MEM_RELEASE; \
-                                        } while (0)
+# define QTHREAD_FASTLOCK_INIT(x)     { atomic_init(&x, false); }
+# define QTHREAD_FASTLOCK_INIT_PTR(x) { atomic_init(x, false); }
+# define QTHREAD_FASTLOCK_LOCK(x)     while ( atomic_load(x) || atomic_exchange_explicit(x, true, memory_order_acquire)) { SPINLOCK_BODY(); }
+# define QTHREAD_FASTLOCK_UNLOCK(x)   do { atomic_store_explicit(x, false, memory_order_release); } while (0)
 # define QTHREAD_FASTLOCK_DESTROY(x)
 # define QTHREAD_FASTLOCK_DESTROY_PTR(x)
 # define QTHREAD_FASTLOCK_TYPE        qt_spin_exclusive_t
-# define QTHREAD_FASTLOCK_INITIALIZER (qt_spin_exclusive_t) {0, 0 }
+# define QTHREAD_FASTLOCK_INITIALIZER false
 #elif defined(HAVE_PTHREAD_SPIN_INIT) && !defined(QTHREAD_OVERSUBSCRIPTION)
 # include <pthread.h>
 # define QTHREAD_FASTLOCK_ATTRVAR
```

include/qt_atomics.h
```diff
@@ -54,9 +67,12 @@ void qt_spin_exclusive_unlock(qt_spin_exclusive_t *);
 # define QTHREAD_FASTLOCK_INIT(x)     { (x).enter = 0; (x).exit = 0; }
 # define QTHREAD_FASTLOCK_INIT_PTR(x) { (x)->enter = 0; (x)->exit = 0; }
 # define QTHREAD_FASTLOCK_LOCK(x)     { aligned_t val = qthread_incr(&(x)->enter, 1); \
-                                        while (val != (x)->exit) SPINLOCK_BODY(); /* spin waiting for my turn */ }
+                                        while (val != (x)->exit) SPINLOCK_BODY(); \
+                                          THREAD_FENCE_MEM_ACQUIRE; /* spin waiting for my turn */ }
 # define QTHREAD_FASTLOCK_UNLOCK(x)   do { COMPILER_FENCE; \
-                                           (x)->exit++; /* allow next guy's turn */ } while (0)
+                                           (x)->exit++; /* allow next guy's turn */ \
+                                           THREAD_FENCE_MEM_RELEASE; \
```
ronawho, line 74: Doesn't the release fence need to be before the code that actually releases the lock? As-is it seems like a thread could execute `(x)->exit++;`, get interrupted and have another thread run and acquire the lock before any pending operations from the original thread are fenced.

janciesko: Yes.

janciesko: Yes, seem wrong here. Even without interruption. Another thread can be polling the while-loop on trying to enter while the thread that has entered the critical section is updating the exit++, thus allowing the polling thread to enter without the exiting thread to call `THREAD_FENCE_MEM_RELEASE`.

ronawho: FYI changing the order allowed 1K trials of my test to pass.

janciesko: Cool, thanks.

include/qt_atomics.h
```diff
@@ -131,10 +147,12 @@ typedef union qt_spin_trylock_s {
 # define QTHREAD_TRYLOCK_INIT(x)     { (x).u = 0; }
 # define QTHREAD_TRYLOCK_INIT_PTR(x) { (x)->u = 0; }
 # define QTHREAD_TRYLOCK_LOCK(x)     { uint32_t val = qthread_incr(&(x)->s.users, 1); \
-                                       while (val != (x)->s.ticket) SPINLOCK_BODY(); /* spin waiting for my turn */ }
-# define QTHREAD_TRYLOCK_UNLOCK(x)   do { COMPILER_FENCE;                                               \
+                                       while (val != (x)->s.ticket) SPINLOCK_BODY(); \
+                                          THREAD_FENCE_MEM_ACQUIRE; /* spin waiting for my turn */ }
+# define QTHREAD_TRYLOCK_UNLOCK(x)   do { COMPILER_FENCE; \
                                           qthread_incr(&(x)->s.ticket, 1); /* allow next guy's turn */ \
```
janciesko, line 153: @ronawho The same ordering issue applies here.

ronawho (approved):
👍   Looks good!

Chapel testing results:
- correctness and performance testing on x86 looks good. 
- correctness testing on ARM passed (tested on ThunderX2 and A64FX)

I can't use my Sandia account at the moment, so didn't test on mayer (also ThunderX2) or weaver (powerpc), but if I remember after I get my access sorted out I'll test then.

cjhackillinois:
What branch was this tested on for Thunder X2 and A64FX?

olivier-snl:
> What branch was this tested on for Thunder X2 and A64FX?

The most up-to-date should be https://github.com/Qthreads/qthreads/tree/PRTestBranchDec2021

ronawho:
I'm just testing this specific PR for now. Will test native context switching later.

olivier-snl:
> I'm just testing this specific PR for now. Will test native context switching later.

So there you go, @cjhackillinois : just https://github.com/janciesko/qthreads/tree/memory_fence_generic

cjhackillinois:
@olivier-snl  This PR doesn't contain the native ARMv8 code. I was just checking to be sure I interpret @ronawho comments correctly.

include/qt_atomics.h
cjhackillinois: Have we decided if we are going to guard with `__STDC_VERSION__` or `__STDC_NO_ATOMICS`

cjhackillinois (approved):
Nothing sticks out to me and should be ready for merge.

include/qt_atomics.h
ronawho: Here's the check we have in our C11 atomic implementation:

```c
#if __STDC_VERSION__ >= 201112L && !defined(__STDC_NO_ATOMICS__)
  #include <stdatomic.h>
  // ...
#endif
```

dmk42: Right, the method quoted by @ronawho is much preferred, since it filters out implementations that are not required to supply `__STDC_NO_ATOMICS__`.

janciesko: Nice. I am adding this to the PR.

janciesko: Done.

PR #100, opened 2021-11-30
Detect Power9 during configure of asm inline support on Lassen

janciesko:
Lassen at LLNL reports `host='powerpc64le-unknown-linux-gnu'` thus the m4 script does not detect the architecture correctly.


olivier-snl (approved):
LGTM.

PR #108, opened 2022-06-06
Increase default stack size from 4kB to 16kB

janciesko:
Proposing to increase the default stack size to 16kB. This allows to use larger stack local variable types and would improve support of many use-cases that segfault otherwise.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -114,7 +114,7 @@ AC_ARG_WITH([default-stack-size],
             [AS_HELP_STRING([--with-default-stack-size=[[bytes]]],
                             [Specify the number of bytes to use for stacks, by
                              default (i.e. when the environment variable
-                             QT_STACK_SIZE is unset). Default: 4096 (4k).
+                             QT_STACK_SIZE is unset). Default: 16384 (16k).
                              Depending on your OS, standard Pthread threads may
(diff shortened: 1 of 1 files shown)


olivier-snl:
Please check on what the default is for Chapel's configuration -- I know it is greater than our current 4k default, but I don't remember what they chose.  It would be good to consider that info in this decision.

janciesko:
Looking at https://chapel-lang.org/docs/usingchapel/executing.html#controlling-the-call-stack-size I see `The default call stack size is ~8 MiB on Linux-based systems, since this is a common value for the process stack limit on such systems. On Cygwin systems the default call stack size is ~2 MiB. ` @ronawho, are there any concerns on your side in respect to this PR?

olivier-snl:
> Looking at https://chapel-lang.org/docs/usingchapel/executing.html#controlling-the-call-stack-size I see `The default call stack size is ~8 MiB on Linux-based systems, since this is a common value for the process stack limit on such systems. On Cygwin systems the default call stack size is ~2 MiB. ` @ronawho, are there any concerns on your side in respect to this PR?

OK.  I don't think we want to go that big for our default.  16kB seems reasonable for a general Qthreads default.

PR #109, opened 2022-06-08
Add html man pages

janciesko:
This adds generated html man files making quick API lookup more convenient.


olivier-snl:
> @olivier-snl Ping

Do these get installed in share along with the existing man pages?  Does this make [htmlize.sh](https://github.com/Qthreads/qthreads/blob/main/man/htmlize.sh) obsolete, and subject to removal?

janciesko:
I think we could leave the gen script. The html files do not install - that's a good point. I'll include them in the make install  step.

janciesko:
For some reason, ./configure does not generate the code for the install target. make install currently does not place the html directory into $(prefix)/man/. I do not know why.

insertinterestingnamehere:
Do we necessarily want to include generated files directly in the git repo?

PR #110, opened 2022-07-19
Enable QTHREAD_ARMV8_A64 context swap on Apple Mx (ARM) hardware

janciesko:
This makes Qthreads compile with fast context switching implemented in asm.

### DIFF
--- a/src/fastcontext/asm.S
+++ b/src/fastcontext/asm.S
@@ -42,8 +42,13 @@
 #  define NEEDPOWERCONTEXT 1
 #  define SET _qt_setmctxt
 #  define GET _qt_getmctxt
+# elif (QTHREAD_ASSEMBLY_ARCH == QTHREAD_ARMV8_A64)
+#  define NEEDARMA64CONTEXT 1
+#  define ISAPPLEMACHOASM 1
+#  define SET _qt_setmctxt
+#  define GET _qt_getmctxt
 # else
-#  error What kind of a Mac is this? If M1, use CFLAGS=-D_XOPEN_SOURCE and --disable-fastcontext --with-default-stack-size=65536 in configure.
+#  error What kind of a Mac is this?
 # endif
 #elif defined(__linux__)
 # if (QTHREAD_ASSEMBLY_ARCH == QTHREAD_ARM)
@@ -793,7 +798,12 @@ SET:
  */
 .globl GET
 .global GET
-.type GET, %function
+#ifndef ISAPPLEMACHOASM
+ .type GET, %function
+#else
+.p2align 4
+#endif
+
 GET:
         /* Store GP registers */
         stp     X1, X2, [X0,#8]
@@ -851,7 +861,11 @@ GET:
 
 .globl SET
 .global SET
-.type SET, %function
+#ifndef ISAPPLEMACHOASM
+ .type SET, %function
+#else
+.p2align 4
+#endif
 SET:
         /* Load GP registers */
         ldp     X1, X2, [X0,#8]


olivier-snl:
Where is ISAPPLEMACHOASM defined?

janciesko:
Make check passes on M1 Mac, x86 Linux, ARM64 Linux.

src/fastcontext/asm.S
```diff
@@ -851,7 +861,11 @@ GET:
 
 .globl SET
 .global SET
-.type SET, %function
+#ifndef ISAPPLEMACHOASM
+ .type GET, %function
```
ronawho, line 865: Seems like this changed what was a SET to a GET?

olivier-snl: That does seem fishy.  @janciesko ?

janciesko: Thanks for reporting. This is wrong. Amended the commit.

ronawho:
I'm seeing failures for a couple types of tests.

One failure for tests that sleep, where it seems like when doing sleeps from different tasks only some of the tasks actually sleep. Our sleep code is implemented [here](https://github.com/chapel-lang/chapel/blob/4c48d6191fa2ea46e2dcb8964a6b2d2327f865a4/runtime/src/tasks/qthreads/tasks-qthreads.c#L950-L983). If I switch to always using the `gettimeofday` code instead of the `qtimer` code then sleeps seem to work correctly. I thought maybe the mach timer initialization in qthreads was racy, but doing a serial timer creation during serial init didn't seem to help

In another mode, multiples tasks adding to a data structure protected by a lock seems to be racy in some way.

I don't have any standalone qthread reproducers yet, but some of the chapel tests are:

```
$CHPL_HOME/test/domains/sungeun/assoc/parSafeMember.chpl
$CHPL_HOME/test/parallel/begin/deitz/test_begin5.chpl
```

These tests pass with fastcontext disabled, and I'm not sure how to go about debugging offhand. Probably the starting point is to get standalone qthread reproducers.

olivier-snl:
> I'm seeing failures for a couple types of tests.
> 
> One failure for tests that sleep, where it seems like when doing sleeps from different tasks only some of the tasks actually sleep. Our sleep code is implemented [here](https://github.com/chapel-lang/chapel/blob/4c48d6191fa2ea46e2dcb8964a6b2d2327f865a4/runtime/src/tasks/qthreads/tasks-qthreads.c#L950-L983). If I switch to always using the `gettimeofday` code instead of the `qtimer` code then sleeps seem to work correctly. I thought maybe the mach timer initialization in qthreads was racy, but doing a serial timer creation during serial init didn't seem to help
> 
> In another mode, multiples tasks adding to a data structure protected by a lock seems to be racy in some way.
> 
> I don't have any standalone qthread reproducers yet, but some of the chapel tests are:
> 
> ```
> $CHPL_HOME/test/domains/sungeun/assoc/parSafeMember.chpl
> $CHPL_HOME/test/parallel/begin/deitz/test_begin5.chpl
> ```
> 
> These tests pass with fastcontext disabled, and I'm not sure how to go about debugging offhand. Probably the starting point is to get standalone qthread reproducers.

@ronawho Do your tests fail only on arm mac or on other platforms as well?  I know you said that they work with the system ucontext.  Do they work with fastcontext on arm linux and x86 mac?  Just trying to narrow down the the problem and figure out if we introduced problems for the other platforms as well since the last round of chapel testing.

ronawho:
These are old tests, so I believe they are only failing for fastcontext on arm based macs. We have lots of x86 testing on mac and linux. We have some arm linux testing, but not as extensive so I can't say confidently at the moment if these tests pass consistently on other arm platforms.

ronawho:
Tests seem to fail on arm (thunderx2) linux with fastcontext. They pass with the system ucontext. So seems to be arm fastcontext specific. I don't know that I would have run these specific tests when checking out https://github.com/Qthreads/qthreads/pull/95. They are part of the full test suite and given that I was mostly testing that PR out on cray's I was probably running a more limited subset of the tests.

olivier-snl:
> Tests seem to fail on arm (thunderx2) linux with fastcontext. They pass with the system ucontext. So seems to be arm fastcontext specific. I don't know that I would have run these specific tests when checking out #95. They are part of the full test suite and given that I was mostly testing that PR out on cray's I was probably running a more limited subset of the tests.

@ronawho Thanks for running the additional testing.  Since the differences between the mac and Linux fastcontext assembly are very minor, it is not surprising that they fail the same tests, and it is helpful to have that confirmation so that we can reproduce and debug on either platform.

ronawho:
Right, and since this is an existing issue and not something new or mac specific, I don't think it needs to hold up this PR, but I think Chapel will want to wait until this is resolved to use native fastcontext switching for ARM.

I'm trying to make a standalone qthreads reproducer from test_begin5.chpl, but wasn't able to in a few minutes. I'm pretty busy the next few days, so not sure when I'll get a chance to try again. This test is kinda silly and could be argued that it's invalid since it uses `sleep` to try and get output in a specific order. However, I think it still exposes an underlying issue and it's a lot easier to create a reproducer that looks like this instead of the other test failure so I'd probably start with that test and hope it's the same underlying issue.

I also tried looking at the argobots arm assembly to see if I could see any obvious bugs in the qthreads version, but they are different enough and my understanding of asm context switching is poor enough that I didn't really come to any conclusion.

janciesko:
@cjhackillinois - would you be able to look at this? We do not catch this error with our testing suite. It might be that some register is not handled correctly and that in this case the register happens to store the state of lock.

janciesko:
@ronawho We will merge this into main. We will keep the issue open that you've reported and will fix that in a minor release (3.8.1).

PR #111, opened 2022-09-08
[WIP] Add support for recursive locks and try_lock 

janciesko:
This implements a recursive lock using FEBs. The idea is that `qthreade_read_FE` becomes re-entrant if `qthread_read_FE` is called from `qthread_lock` and `qthread_lock_init` has been called on the same address setting `is_recursive` to true. Analogously, `qthread_unlock` calls `qthead_fill` which decrements a counter of a recursive lock. If the counter reaches zero, blocked waiters (EFQ) are unblocked. 

Uses of other FEB calls over recursive locks are unsupported and we could error out. Currently they have no effect but could lead to programming errors if programmer expects FEB synchronization over an address used for a recursive lock.

Todo
- [ ] Add unit test
- [ ] Error out in incorrect use
- [ ] Update qthread_feb_blocker_func to use proper func name

With this infrastructure, we should be able to support ULTs from Open MPI. 
For Open MPI, we'd need then these changes:
https://github.com/janciesko/ompi/commit/daef600cee3ed9607f10829c5dc3b4de5da0bc4c
https://github.com/janciesko/ompi/commit/bab853bf53355c6fbe71f24a4321bf19293d1165

### DIFF
--- a/include/qt_addrstat.h
+++ b/include/qt_addrstat.h
@@ -18,6 +18,8 @@ static QINLINE qthread_addrstat_t *qthread_addrstat_new(void)
         ret->FEQ   = NULL;
         ret->FFQ   = NULL;
         ret->FFWQ  = NULL;
+        ret->acq_owner_stat.state = -1;
+        ret->acq_owner_stat.recursive_access_counter = 0;
         QTHREAD_EMPTY_TIMER_INIT(ret);
     }
     return ret;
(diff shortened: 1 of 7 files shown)


janciesko:
This is not going to fly. The problem is that unlike Pthreads, FEBs do not poll a state but generate the dependency graph at instantiation and thus cannot ignore conditions based on ownership (runtime property). At runtime, we'd need to modify the graph in order to ignore prod-cons but the notion of recursiveness is lost as we cannot associate inner producers to inner consumers.

PR #112, opened 2022-09-15
 [WIP] Add support for recursive locks and try_lock (using spinlocks)

janciesko:
This is another implementation using hashmaps of spinlocks and likely the way to go. Needs a bit of work
- [x] unit test
- [x] fix compilation errors
- [x] test on x86
- [x] test on PowerISA 
- [x] test on ARM64
- [x] test on MAC(x86,ARM64)

With this infrastructure, we should be able to support ULTs from Open MPI.
For Open MPI, we'd need then these changes:
https://github.com/janciesko/ompi/commit/daef600cee3ed9607f10829c5dc3b4de5da0bc4c
https://github.com/janciesko/ompi/commit/bab853bf53355c6fbe71f24a4321bf19293d1165

### DIFF
--- a/include/qt_hash.h
+++ b/include/qt_hash.h
@@ -14,10 +14,14 @@
 extern "C" {
 #endif
 
+#define PUT_COLLISION 0
+#define PUT_SUCCESS   1
+
 typedef const void *qt_key_t;
 typedef struct qt_hash_s *qt_hash;
 typedef void (*qt_hash_callback_fn)(const qt_key_t, void *, void *);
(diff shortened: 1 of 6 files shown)


olivier-snl (commented):
I think this is okay.  We should be able to afford the time for the one conditional in the base case without much overhead penalty.

janciesko:
`./test/stress/lock_scq_rel` passes on all supported architectures.

janciesko:
` ake check` passes on all supported architectures.

PR #118, opened 2022-09-20
Improve distrib gen script

janciesko:
- Adds separators
- Replaces -j option with -z (bzip2 vs gzip)

### DIFF
--- a/scripts/package.sh
+++ b/scripts/package.sh
@@ -1,4 +1,16 @@
 #!/bin/bash
+
+# Gen script for distributable package
+# Usage: cd SRC_DIR/scripts && bash package.sh TAG
+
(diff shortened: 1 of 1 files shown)


janciesko:
This script generates the correct gzipped tarball. It accepts a tag as input argument. GitHub Automation could trigger this script and simplify the release procedure, however, running this manually and adding that file manually is fairly easy.

olivier-snl (commented):
Looks handy. I'm generally going to be okay with any of these sorts of changes since the packaging automation files are really for our use and not the end users.

janciesko:
Feel free to click the Approve button, so we'd follow protocol :)

olivier-snl (approved):

PR #119, opened 2022-09-20
Add support for recursive locks with external lock allocation

janciesko:
This adds support for locking with external lock allocations. In this case, the external lock must be a qthread_spinlock_t thus the API has the name `spinlock` to disambiguate. This covers the use-case coming from ompi where ompi provides the lock.

### DIFF
--- a/include/qt_atomics.h
+++ b/include/qt_atomics.h
@@ -144,13 +144,6 @@ extern pthread_mutexattr_t _fastlock_attr;
     && defined(QTHREAD_ATOMIC_INCR)                           \
     && !defined(QTHREAD_MUTEX_INCREMENT)
 
-typedef union qt_spin_trylock_s {
-    aligned_t u;
-    struct {
-        haligned_t ticket;
-        haligned_t users;
-    } s;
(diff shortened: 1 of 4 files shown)


olivier-snl (commented):
Noticed that this is using C bool type and its header, which are C99 and above. Do we need to do something for <C99 -- guard or just used unsigned int instead?

janciesko:
According to our autoconf script (`./configure.ac`): 
```
if test "x$ac_cv_prog_cc_c99" == "xno"; then :
  as_fn_error $? "Qthreads requires a C compiler that supports the C99 standard." "$LINENO" 5
fi
ac_ext=c

janciesko:
I think we indeed need C99 as we rely on variadic functions etc.

olivier-snl:
> According to our autoconf script (`./configure.ac`):
> 
> ```
> if test "x$ac_cv_prog_cc_c99" == "xno"; then :
>   as_fn_error $? "Qthreads requires a C compiler that supports the C99 standard." "$LINENO" 5
> fi
> ac_ext=c
> ```

OK.  Yes, that seems to prove your point.  C99 is a fine baseline.  We'll still keep guarding for C11, but relevant to the current discussion you're use of C99 bool should be fine then.

olivier-snl (approved):

PR #120, opened 2022-09-21
Improve version format and update version number

janciesko:
- Improves version format and updates version number to current.

(I would prefer a version file (version.in) and a version.h header where configure sets a string. This would give a cleaner version history. We can do that someday. )


olivier-snl (commented):
Certainly more robust than what we had before.  Do we know if any of our users (Chapel, RaftLib, etc.) check this number currently.  I would guess perhaps not since it hadn't been updated in some time.

olivier-snl (commented):
Comment deleted -- was regarding other PR

olivier-snl (approved):

PR #121, opened 2022-11-09
Make stack size in number of word-size entries for ARM64

janciesko:
@cjhackillinois - you had this change in  mind?
`make check` passes on ARM64.


cjhackillinois (approved):
Yes, this looks correct.

PR #122, opened 2022-11-10
Add return value in inner function in stackleft unit test

janciesko:
-O2 seems to override Q_NOINLINE thus inner recursions would not advance stack ptr. A return value enforces this. 
I have uncommented the commented section which now works.


olivier-snl (approved):

cjhackillinois:
Should the test work regardless of in-lining?

janciesko:
No, I don't think so.

Issue #123, opened 2022-11-23
Fast context swap defective in some cases on ARM64


janciesko:
Scenario 2 in the code below fails to produce the correct output sequence when not using --disable-fastcontext on ARM64.
Reproduces issue reported in [#113](https://github.com/Qthreads/qthreads/issues/113).

```C
#ifdef HAVE_CONFIG_H
# include <config.h>
#endif

#include <stdio.h>
#include <stdlib.h>
#include <assert.h>
#include <unistd.h>
#include <qthread/qthread.h>
#include "qthread/qtimer.h"

aligned_t cond[2];

#define taskwait \
    status &= qthread_readFE(NULL, &cond[0]); \
    status &= qthread_readFE(NULL, &cond[1]);

#define reset \
    status &= qthread_empty(&cond[0]); \
    status &= qthread_empty(&cond[1]);

double t_long = 2.0;
double t_short = 1.0;

aligned_t task_long(void *arg) {
    double secs = t_long;
    qtimer_t t = qtimer_create();
    qtimer_start(t);
    do {
        qthread_yield();
        qtimer_stop(t);
    } while (qtimer_secs(t) < secs);
    printf("---\n");
    qthread_writeF(&cond[0], &cond[0]);
    return 0;
}

aligned_t task_short(void *arg) {
    double secs = t_short;
    qtimer_t t = qtimer_create();
    qtimer_start(t);
    do {
        qthread_yield();
        qtimer_stop(t);
    } while (qtimer_secs(t) < secs);
    printf("17\n");
    qthread_writeF(&cond[1], &cond[1]);
    return 0;
}

int scenario_1() {
    int status = QTHREAD_SUCCESS;   
    reset;
    printf("---\n");
    status &= qthread_fork(task_long, NULL, NULL);    
    status &= qthread_fork(task_short, NULL, NULL);
    taskwait;
    return status;
}

aligned_t task_short_inner(void *arg) {
    double secs = t_short;
    qtimer_t t = qtimer_create();
    qtimer_start(t);
    do {
        qthread_yield();
        qtimer_stop(t);
    } while (qtimer_secs(t) < secs);
    printf("17\n");
    qthread_writeF(&cond[1], &cond[1]);
    return 0;
}

aligned_t task_long_outer(void *arg) {
    int status = QTHREAD_SUCCESS;
    printf("---\n");
    status &= qthread_fork(task_short_inner, NULL, NULL);
    assert(status == QTHREAD_SUCCESS);
[...]

Issue #113, opened 2022-09-16
Some Chapel tests fail on Apple Mx hw and fast context swap on


janciesko:
As reported by @ronawho 

> I'm seeing failures for a couple types of tests.
> 
> One failure for tests that sleep, [here](https://github.com/chapel-lang/chapel/blob/4c48d6191fa2ea46e2dcb8964a6b2d2327f865a4/runtime/src/tasks/qthreads/tasks-qthreads.c#L950-L983) it seems like when doing sleeps from different tasks only some of the tasks actually sleep. Our sleep code is implemented here. If I switch to always using the gettimeofday code instead of the qtimer code then sleeps seem to work correctly. I thought maybe the mach timer initialization in qthreads was racy, but doing a serial timer creation during serial init didn't seem to help
> 
> In another mode, multiples tasks adding to a data structure protected by a lock seems to be racy in some way.
> 
> I don't have any standalone qthread reproducers yet, but some of the chapel tests are:
> 
> $CHPL_HOME/test/domains/sungeun/assoc/parSafeMember.chpl
> $CHPL_HOME/test/parallel/begin/deitz/test_begin5.chpl
> 
> These tests pass with fastcontext disabled, and I'm not sure how to go about debugging offhand. Probably the starting point is to get standalone qthread reproducers.

Further discussion:
https://github.com/Qthreads/qthreads/pull/110

janciesko:
Fixed in https://github.com/Qthreads/qthreads/pull/125

ronawho:
Nope, resolved by #125

PR #125, opened 2023-04-14
FPU store/load for arm64 fastcontext

stelleg:
This addresses issue #123 by storing/loading FPU registers in the arm64 fastcontext.

### DIFF
--- a/include/fastcontext/arm-ucontext.h
+++ b/include/fastcontext/arm-ucontext.h
@@ -20,9 +20,7 @@ struct mctxt {
     /* Saved main processor registers. */
 #ifdef NEEDARMA64CONTEXT
 	uint64_t regs[32]; /* callee saves x0-x30, SP */
-	#ifdef ARMA64_CONTEXT_SWITCH_NEON_REGS
-	uint128_t regs[32]; /* SIMD Neon Registers*/
-	#endif
+	uint64_t fpu_regs[64]; /* 32 128bit FPU/SIMD Neon Registers */
 #else
     uint32_t regs[16]; /* callee saves r0-r15 */
 #endif
--- a/src/fastcontext/asm.S
+++ b/src/fastcontext/asm.S
@@ -821,12 +821,11 @@ GET:
         stp     X25, X26, [X0,#200]  
         stp     X27, X28, [X0,#216]
         stp     X29, X30, [X0,#232]     /* X30 is LR */
-	mov	X29, SP
+        mov     X29, SP
         str     X29, [X0,#248]           /* sp, not sure if ARMv8 allows storing SP directly */
-	ldr	X29, [X0,#232]		/* restore FP just to be safe */
+        ldr     X29, [X0,#232]           /* restore FP just to be safe */
 
-#ifdef ARMA64_CONTEXT_SWITCH_NEON_REGS
-        /* Store vector registers */
+        /* Store fpu/vector registers */
         add     X0, X0, #256            /* Update pointer to avoid immediate growing to large for STP/LDP instruction */
         stp     Q0, Q1, [X0]
         stp     Q2, Q3, [X0,#32]
@@ -845,14 +844,14 @@ GET:
         stp     Q28, Q29, [X0,#448]
         stp     Q30, Q31, [X0,#480]
         sub     X0, X0, #256
-#endif
+
         /* Store 1 as X0-to-restore */
         mov     X1, #1
         str     X1, [X0]                
         
         /* return 0 */
         mov     X0, #0
-	ret
+        ret
 
 /*
  * TODO: add vector registers for both NEON and SVE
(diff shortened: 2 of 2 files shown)


janciesko:
CI testing is failing likely due to https://github.com/Qthreads/qthreads/issues/126.

olivier-snl:
It's great to have a potential fix to the bug.  We'll want to do broad testing.  In particular I'm interested to see if we need to do anything differently to handle ThunderX2 vs A64FX vs arm mac.  The reproducer derived from the failed chapel test happens to have a double time value.  I think a good unit test should be constructed to have some intentional FP work (including vector FP), then yield, then more FP.

include/fastcontext/arm-ucontext.h
```diff
@@ -19,7 +19,7 @@ typedef struct uctxt uctxt_t;
 struct mctxt {
     /* Saved main processor registers. */
 #ifdef NEEDARMA64CONTEXT
-	uint64_t regs[32]; /* callee saves x0-x30, SP */
+	uint64_t regs[40]; /* callee saves x0-x30, SP, FPU:d8-d16 */
```
cjhackillinois, line 22: Why 8 FPU registers? There are a total of 32 for Aarch64, no?

src/fastcontext/asm.S
```diff
@@ -883,12 +889,18 @@ SET:
         ldp     X25, X26, [X0,#200]  
         ldp     X27, X28, [X0,#216]
         ldr     X29, [X0,#248]
-	mov	SP, X29
-	ldp	X29, X30, [X0,#232]
+        mov     SP, X29
+        ldp     X29, X30, [X0,#232]
+
+        // FPU
+        ldp     D8,  D9,  [X0, #264]
```
cjhackillinois, line 896: Why start at offset of 264 instead of 256? 32 8-bytes registers is 256 bytes, so I would assume you could start saving the D registers at offset 256. Also, why these 8 registers?

cjhackillinois (commented):
I would like to see a reference why only 8 FPU registers are saved out of the 32? If a double is used, does gcc put the double into D8 as opposed to D0 or D1?

stelleg:
Some responses and updates to the helpful feedback from @cjhackillinois:

So the changes I had added above were borrowed from here: https://github.com/pmodels/argobots/blob/main/src/arch/fcontext/fcontext_arm64_aapcs_elf_gas.S. I thought maybe they knew something we didn't, but upon further inspection, I think the existing code that was ifdef'd out as NEON_REGS is doing the more complete FPU context swap. I've reverted my changes and replaced them with a simple removal of the NEON_REGS ifdef and a minor buffer change, because as I understand it those are non-optional in ARM64 and required for floating point, so I don't think it makes sense to have them be ifdef'd out.

janciesko:
@stelleg - Could you please rebase your PR onto the current main branch? That should make CI testing pass.

include/fastcontext/arm-ucontext.h
```diff
@@ -20,9 +20,7 @@ struct mctxt {
     /* Saved main processor registers. */
 #ifdef NEEDARMA64CONTEXT
 	uint64_t regs[32]; /* callee saves x0-x30, SP */
-	#ifdef ARMA64_CONTEXT_SWITCH_NEON_REGS
-	uint128_t regs[32]; /* SIMD Neon Registers*/
-	#endif
+	uint64_t fpu_regs[64]; /* 32 128bit FPU/SIMD Neon Registers */
```
cjhackillinois, line 23: I think this could use the uint128_t type.

stelleg: My understanding is that uint128_t is not required by the c specification, and often isn't defined. It wasn't for the compiler/math.h I was using when developing.

dmk42: > My understanding is that uint128_t is not required by the c specification, and often isn't defined. It wasn't for the compiler/stdint.h I was using when developing.

That is correct.  Only up to uint64_t is required by the standard (and only if the target supports it).  The Mac is one of the reasons for this PR, and it does not have uint128_t.

dmk42: FYI, C23 will require exact-width types for any width supported by the implementation, so in the future, if a 128-bit integer type is available, then uint128_t will be required.  C23 has not been published yet, though.

cjhackillinois (approved):

include/fastcontext/arm-ucontext.h
cjhackillinois: I believe I tested with "qthread/qthread-int.h". That header file is generated by `./configure`.

ronawho:
Wohoo, very excited to see this! Both of the Chapel tests noted in https://github.com/Qthreads/qthreads/issues/113 that were failing do use floating points. I have confirmed those now run correctly on thunderx2.

If there's interest, I could run more comprehensive Chapel testing on thunderx2, a64fx, m1, and graviton.

Could also see if there's much difference in task creation/switching speed before and after this if that's a concern.

olivier-snl:
@ronawho on the mac side it seems that there may have been recent apple changes that will mean a little additional fix on that platform.  @janciesko reported that this may be Ventura-related.

We don't have access to graviton (that I know of).

ronawho:
> @ronawho on the mac side it seems that there may have been recent apple changes that will mean a little additional fix on that platform. @janciesko reported that this may be Ventura-related.

Just curious, Is that with r17 not being available or something else?

> We don't have access to graviton (that I know of).

I meant, I have access to those 4 platforms if you'd like me to run some testing on any of them.

janciesko:
On Ventura, it seems that autoconf is not correctly detecting AArch64 and thus sets `asssembly architecture` to  `UNSUPPORTED`.

PR #132, opened 2023-05-24
Fix --disable-hwloc-configure-checks behavior

jhh67:
I botched my last PR #131, it didn't set `CPPFLAGS` and `LDFLAGS` properly if `hwloc` link checks were disabled. I think I forgot to run `autoreconf` at some point during testing so I was testing against an out-of-date `configure` script.


janciesko:
Great, thanks for the hotfix John.

PR #133, opened 2023-06-13
Get topology from external function

jhh67:
If `HWLOC_GET_TOPOLOGY_FUNCTION` is defined, call it to get the `hwloc` topology instead of calling `hwloc_topology_init` and `hwloc_topology_load`. This enables `qthreads` to share the same `hwloc` topology with external code, and obviates `hwloc_via_chapel` affinity. The `configure` script accepts the argument `--with-hwloc-get-topology-function` to set `HWLOC_GET_TOPOLOGY_FUNCTION`.


olivier-snl:
There are some failed CI checks on this.  We'll need to look into whether they are actually related to the changes in the PR.

janciesko:
@jhh67 - Could you please consider resetting `topology` to `NULL` in the hwloc teardown? The problem is that a sequence of `qthreads_init` and `qthread_finalize` preserves the global variable and thus we do not reenter the branch to repopulate data structures via `hwloc_topology_init` in a 2nd sequence. Alternatively, do not check for NULL?

janciesko:
Using binders, this would still fail?

jhh67:
I pushed a commit that fixes topology cleanup in the `hwloc` teardown routine. I tested it on the `hello_world` and `reinitialization` tests, I don't know how to run all the tests and verify the results (give me a pointer on how to do that and I'll run all the tests). I'm not sure about `binders` -- it doesn't have a teardown routine. It should just get the topology each time it is initialized.

src/affinity/binders.c
```diff
@@ -91,9 +91,15 @@ int INTERNAL qt_affinity_compact(int num_workers,
...
                                size_t                *hw_par)
-{                           
-  hwloc_topology_init(&topology);
-  hwloc_topology_load(topology);
+{
+#ifdef HWLOC_GET_TOPOLOGY_FUNCTION
+  extern void * HWLOC_GET_TOPOLOGY_FUNCTION;
+  topology = (hwloc_topology_t) HWLOC_GET_TOPOLOGY_FUNCTION;
+#endif
+  if (topology == NULL) {
```
janciesko, line 99: What about removing this conditional

janciesko: I see your point.

janciesko: @jhh67 Could you add a comment into the code that the lack of a tear_down routine makes qt_affinity_init skip topology initialization in case of qthreads reinitialization? If we update binders at some point we'd likely need to add a tear_down routine.

ronawho:
Note that there's leftover `hwloc_via_chapel` references in `configure.ac` and `src/Makefile.am`

src/affinity/binders.c
jhh67: Done.

janciesko (approved):

janciesko:
LGTM

PR #136, opened 2023-08-21
Add a mechanism to reset the default task spawn order

ronawho:
Add `qthread_reset_target_shep()`, which resets the shepherd that tasks created with `NO_SHEPHERD` get spawned to. This is useful in Chapel for cases where we want to use the default spawn location, but get good NUMA affinity between data parallel loops sepreated by some other task spawn.

Just using the `qthread_fork_to` API would disable future stealing, so this is a mechanism  to get tasks in consecutive data parallel loops to spawn to the same shepherds without preventing stealing.

This was originally done in chapel-lang/chapel#12868, which includes some perf results. At the time I expected we'd update our shim to use `qthread_fork_to`, but had not considered the work-stealing implications at the time.

### DIFF
--- a/include/qthread/qthread.h
+++ b/include/qthread/qthread.h
@@ -333,6 +333,9 @@ int qthread_spawn(qthread_f             f,
 /* This is a function to move a thread from one shepherd to another. */
 int qthread_migrate_to(const qthread_shepherd_id_t shepherd);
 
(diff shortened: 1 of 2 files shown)


janciesko:
LGTM. Thanks! I will add documentation and an entry to the 1.19 release log shortly.

janciesko (approved):

PR #137, opened 2023-08-28
Update NEWS and version number for 1.19 release

janciesko:


olivier-snl:
In earlier versions we don't seem to have mentioned arm linux support, though we did mention arm mac.  The text for this release says we fixed the context swap bug, which is true, but I think we should also basically make some statement about arm linux and mac being what we consider mature at this stage.  We may also want to add an informational mention of the repo move, just so it is also captured in the NEWS file itself.

olivier-snl (approved):

PR #141, opened 2023-10-17
Add test binaries to .gitignore.

insertinterestingnamehere:
This adds some additional build artifacts to the .gitignore. In particular it helps me get a clean build environment via `git clean -fdX` (note the capital X which only clears away ignored files).

### DIFF
--- a/test/basics/.gitignore
+++ b/test/basics/.gitignore
@@ -17,11 +17,14 @@ qthread_dincr
 qthread_disable_shepherd
 qthread_fincr
 qthread_fork_precond
+qthread_fp
+qthread_fp_double
 qthread_id
 qthread_incr
 qthread_migrate_to
 qthread_readstate
 qthread_stackleft
+qthread_timer_wait
 qtimer
 queue
 read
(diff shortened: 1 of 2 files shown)


janciesko:
Thanks! Why only those specific binaries. Are we messing up the Makefile so make clean won't remove those?

insertinterestingnamehere:
Offhand, I'm not actually sure what's going on with the others. I was using git to clean out the build artifacts, not make, but these were the only ones that weren't getting removed. Maybe the others are already covered by some other exclusion rule elsewhere. I'll check and make sure we're handling things consistently.

insertinterestingnamehere:
The other tests were ignored in the local gitignore files so I switched this PR update those appropriately.

insertinterestingnamehere:
Running `make clean` isn't as thorough of a purge of files as I was looking for. On the other hand, `make distclean` and `make maintainer-clean` still miss a bunch of stuff that `git clean -fdX` picks up. I'm not familiar enough with autoconf to know why that is though. I'll make an issue and come back to it.

janciesko:
Sounds good, yes, I think those are indeed missing in .gitignore

janciesko (approved):

insertinterestingnamehere:
@olivier-snl I don't have merge access, can you merge this one?

PR #170, opened 2023-11-02
Fix Warnings

insertinterestingnamehere:


insertinterestingnamehere:
Okay, this one's ready to go. It resolves all warnings enabled in the "picky" build except for those involved in https://github.com/sandialabs/qthreads/issues/173. Once that's resolved too we can enable -Werror to avoid regressions here. I haven't looked at anything from `-Wextra` yet.

src/fastcontext/context.c
```diff
@@ -60,7 +71,9 @@ void INTERNAL qt_makectxt(uctxt_t *ucp,
 #if 0
     memcpy(sp, &argc + 1, argc * sizeof(uintptr_t));
 #else
+QT_SUPPRESS_BOUNDS_CHECK_WARNINGS
     *(uintptr_t*)sp = *(uintptr_t*)((&argc)+1);
```
insertinterestingnamehere, line 75: Most of these changes are straightforward. This one's unusual though. I wasn't able to figure out how to get around the warning since it's technically correct to warn about an out-of-bounds access here. As far as I can tell though, there's just not a better way to get the address of the varargs than what this is already doing so instead I suppressed the warning.

insertinterestingnamehere:
Worth noting: this doesn't fix all the warnings in the config scripts since a lot of those warnings are inherited from stuff happening upstream with autotools. I just fixed the warnings internal to the configure scripts that have obvious fixes and come from files in qthreads that aren't vendored in from somewhere else.

src/fastcontext/context.c
insertinterestingnamehere: Actually, thinking about this some more, it's not actually forwarding the address of the varargs, just the value, so this *can* actually be done without the out of bounds access. I'll fix that.

insertinterestingnamehere: Okay, getting access to the varargs through the standard mechanisms appears to have worked fine for resolving that warning so it no longer needs to be suppressed and this is good to go.

src/envariables.c
```diff
@@ -87,9 +87,9 @@ unsigned long INTERNAL qt_internal_get_env_num(const char   *envariable,
     unsigned long tmp = dflt;
 
     if (dflt != 0) {
-        char dflt_str[10];
+        char dflt_str[21];
 
-        snprintf(dflt_str, 10, "%lu", dflt);
+        snprintf(dflt_str, 21, "%lu", dflt);
```
janciesko, line 92: Why 21?

janciesko (changes requested):
Regarding replacing the `INTERNAL` macro - is the function attribute not valid for all compilers? If we would like to restrict the visibility to hidden for dynamic linkage via `static`, we might consider changing the macro implementation instead.

src/fastcontext/context.c
```diff
@@ -71,7 +65,7 @@ void INTERNAL qt_makectxt(uctxt_t *ucp,
     /* HOWEVER, the function may not be expecting to pull from the stack,
      * several 64-bit architectures expect that args will be in the correct
      * registers! */
-    ucp->mc.mc_edi = va_arg(argp, uintptr_t);
+    ucp->mc.mc_edi = *(uintptr_t*)sp;
```
janciesko, line 68: This makes more sense to me. Thanks!

src/envariables.c
insertinterestingnamehere: The max length of an unsigned long is 20 digits. `snprintf` needs one more character than the max size, so 21.

insertinterestingnamehere:
> Regarding replacing the `INTERNAL` macro - is the function attribute not valid for all compilers? If we would like to restrict the visibility to hidden for dynamic linkage via `static`, we might consider changing the macro implementation instead.

I'm not actually changing the `INTERNAL` macro here. What I did do is move some functions that were declared as `INTERNAL` to being static instead since they weren't used anywhere else. In a lot of the cases this was the simplest way to clear up the missing prototypes warnings.

insertinterestingnamehere:
To clarify a bit more: there are a few places the `INTERNAL` macro is still needed and used. I don't see the need to change it. It's what's needed for specifying functions that we want to be internal to our `.so` but that need to be linked across translation units. The functions I changed to static aren't used across translation units though so making them static made more sense to get rid of the function prototype warnings (as opposed to doing some kind of forward declaration that allows them to be used elsewhere).

janciesko:
Ok, thanks, in that case, would it make sense to add a new macro, maybe `FUNCTION_IS_UNUSED` that expands to the `static` qualifier and using that in those functions?

insertinterestingnamehere:
I think `static` really is the correct thing here. These functions are used, but only within their translation unit. If they were completely unused I'd just remove them entirely. In theory we could do a macro for translation-unit local linkage, but the extra layer of indirection doesn't really give us anything in this case. Across C implementations that's consistently what `static` does for functions anyway.

janciesko:
Agreed

PR #171, opened 2023-11-02
CI for Intel Compilers

insertinterestingnamehere:


insertinterestingnamehere:
This one's ready to go. One note: I removed the clang 17 and gcc 13 builds from the build matrix that covers compiler versions because there are equivalent builds happening in the linux thorough build that uses only the latest compilers.

olivier-snl (approved):

PR #176, opened 2023-11-07
Stackleft Debugging

insertinterestingnamehere:


insertinterestingnamehere:
Interesting. Okay, I think what was going on with stackleft was two things:
- the noinline feature detection code was broken (fixed in https://github.com/sandialabs/qthreads/pull/170)
- one of the other optimization passes in clang was causing something similar to inlining to happen even when noinline is applied. Supposedly this can happen (see https://stackoverflow.com/a/54482070).

Now ideally I'd like to get to a point where this test doesn't require suppressing compiler optimizations to succeed, but this at least achieves parity between gcc and clang WRT https://github.com/sandialabs/qthreads/issues/144.

olivier-snl (commented):
As this is an evolving set of changes as we debug this one, perhaps we should continue on a branch.  Why not bring your branch over into the sandialabs repo from your private fork and continue here.  Unless you think that rebasing to pull in the changes you are working on for other issues will be a pain?

insertinterestingnamehere:
I'd prefer to keep everything in the main branch if possible. This fix decreases the urgency of the stackleft bug a bit so I can focus on other things and then come back to it. It substantially decreases the noise in CI too which will help while working on other stuff.

PR #179, opened 2023-11-10
Debugging MUSL Libc

insertinterestingnamehere:
Currently just testing in CI to check whether https://github.com/sandialabs/qthreads/issues/160 is easy to fix or not. I'll mark this as ready and request review if the fix does end up being simple.

Depends on https://github.com/sandialabs/qthreads/pull/170.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,30 @@ jobs:
               timeout --foreground -k 10s 4m make check
           no_output_timeout: 30s
 
+  musl:
+    parameters:
+      worker_type:
+        type: string
+      scheduler:
+        type: string
+      topology:
+        type: string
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Interesting. Actually an extremely minor change in the configure check is all that's needed to get qthreads working on MUSL. There are some bugs that show up in the MUSL builds, but they've all already been documented with other configurations (https://github.com/sandialabs/qthreads/issues/146, https://github.com/sandialabs/qthreads/issues/150, https://github.com/sandialabs/qthreads/issues/164). Given that, this is good to go. Instead of having this depend on https://github.com/sandialabs/qthreads/pull/170, I'll patch up the history to eliminate the dependency.

Issue #150, opened 2023-10-25
sync_workers Hangs Intermittently on Clang/Linux


insertinterestingnamehere:
Noticed this in CI. sync_workers occasionally hangs when building with Clang on Linux. This may be related to https://github.com/sandialabs/qthreads/issues/145, but I'm creating a separate issue for this until that can be confirmed.

insertinterestingnamehere:
Same with syncvar_prodcons.

insertinterestingnamehere:
Same with sync_null

insertinterestingnamehere:
Same with aligned_prodcons.

insertinterestingnamehere:
Actually these failures and those documented in https://github.com/sandialabs/qthreads/issues/151 and https://github.com/sandialabs/qthreads/issues/153 all only show up with clang/Linux/sherwood scheduler when using 'hwloc' or 'binders' topology. Given that the circumstances are so specific it seems like these are probably all related. I'll close those issues in favor of this one.

Interestingly enough, https://github.com/sandialabs/qthreads/issues/145 is not isolated to these circumstances so I'll keep that one separate.

This means that the tests affected (that I've found thus far) are:
```
hello_world
qthread_incr
qthread_dincr
arbitrary_blocking_operation
sync_workers
syncvar_prodcons
sync_null
aligned_prodcons
```

This list is likely not exhaustive since many of these failures are very intermittent, but it's a good starting point.

insertinterestingnamehere:
Interestingly, I am seeing the sync_workers hang with the nemesis scheduler when the "paranoia" checks are enabled, so maybe there's some more info there too.

insertinterestingnamehere:
Note: enabling assertions makes these bugs disappear for some reason.

insertinterestingnamehere:
Just saw qthread_incr hang on OSX/M1. This is with the Nemesis scheduler. Probably related, but it might still be a separate bug.

insertinterestingnamehere:
I think I just fixed the sync_workers failure while debugging an undefined sanitizer error (that or it just became dramatically more infrequent). As far as I can tell it was just caused by some bad synchronization in the test itself that was allowing garbage values into the computation and allowing it to sometimes hang instead of terminating correctly. This would mean that the other intermittent hangs in other tests may not be caused by the same code internal to the library and instead be caused by similar API misuse in those tests too.

The consistent failure on ARM documented in https://github.com/sandialabs/qthreads/issues/164 is still present.

[...]

Issue #145, opened 2023-10-18
Intermittent fork_precond failures on Clang/Linux/x86_64


insertinterestingnamehere:
With clang versions 9 through 17 on x86_64 Linux the qthread_fork_precond test sometimes hangs and sometimes completes successfully.

I'll add more details here as I find them.

insertinterestingnamehere:
I managed to reproduce this with gcc 13 as well, but thus far only when clang is installed. I'll see if it appears after a few more tries with gcc or if there's any other difference in the build environment that could cause this.

insertinterestingnamehere:
With gcc (and clang present), the stackleft test still succeeds, but there's also an intermittent hang in qswsrqueue as well that shows up sometimes if fork_precond succeeds.

insertinterestingnamehere:
Stackleft failure is specifically in the "single consumer, multiple producers (array)" portion of the test.

I'm having trouble consistently reproducing the qswsrqueue hang again which is disconcerting, but for now I'll set it it aside.

insertinterestingnamehere:
This is present with OSX/M1 as well, so it's likely an issue with clang.

insertinterestingnamehere:
This is resolved by one of the sanitizer patches.

PR #181, opened 2023-11-13
Relative Error Checking for Floating Point Tests. Fixes #172.

### LINKED ISSUE #172: qthread_fp Fails with icx
There's a consistent test failure in qthread_fp when building with icx.

insertinterestingnamehere:
If a compiler is optimizing overly aggressively it may make assumptions about commutativity/associativity of floating point numbers that don't actually hold perfectly in practice. That's a likely cause of the current test failure with icx. The result in the test is just slightly off from the expected value. While this could arguably be a bug upstream, it's generally a bad idea to use exact float equality for tests. Relative error is a more reliable measure. This change makes the float-related tests check relative error rather than actual equality. The thresholds are very conservative, but checking this way should insulate our test suite from failures resulting in small differences in floating point handling at the compiler and/or hardware level.

### DIFF
--- a/test/basics/qthread_fp.c
+++ b/test/basics/qthread_fp.c
@@ -2,6 +2,7 @@
 #include <config.h>
 #endif
 
+#include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


janciesko:
LGTM

janciesko (approved):

janciesko:
I see some of the CI tests failing - is that fixed in #182 ?

insertinterestingnamehere:
The CI failures include documented but not yet fixed errors like #145, #150, and various others. Icx in particular is based on clang, so it's susceptible to most of the same bugs. This PR covers the only icx-specific issue. Any remaining icx failures also apply to clang. Generally, all the CI failures should be documented and I'm currently working through fixing them all. I figured it would be better to have a thorough but partially passing CI suite because it still lets me see regressions in partially-working configs while debugging.

#181 just covers all errors from the undefined behavior sanitizer. I also happened to pick up part of #150 while resolving those. The lingering issues in the undefined sanitizer builds there are either specific to ARM or specific to clang instead of being specific to the undefined behavior sanitizer.

PR #182, opened 2023-11-14
Fix Undefined Behavior Sanitizer Errors

insertinterestingnamehere:
Depends on both https://github.com/sandialabs/qthreads/pull/181 and https://github.com/sandialabs/qthreads/pull/170. I'll revise it to not depend on those if that ends up being possible after debugging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
           make -j2
       - run:
           command: timeout --foreground -k 10s 2m make check
-          no_output_timeout: 30s
+          no_output_timeout: 60s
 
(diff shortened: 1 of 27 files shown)


insertinterestingnamehere:
Okay, this one is ready to go too. I don't want to try to untangle it from #170 and #181 though, so those need to be reviewed/merged first.

janciesko:
Great, maybe rebase this once #170 is in so I see only the new stuff.

insertinterestingnamehere:
Rebased. Thanks!

janciesko (approved):
LGTM

PR #185, opened 2023-11-28
Have Address Sanitizer Check Init Order In CI

insertinterestingnamehere:
This is a minor CI improvement, but it does depend on #182.


janciesko:
It seems that this PR includes commits from other PRs.

PR #186, opened 2023-12-01
Resolve Race Condition in Nemesis Threadqueue (#187)

insertinterestingnamehere:
Depends on #185 and (transitively) #182.

### DIFF
--- a/include/qt_atomics.h
+++ b/include/qt_atomics.h
@@ -1,23 +1,14 @@
 #ifndef QT_ATOMICS_H
 #define QT_ATOMICS_H
 
+#include <stdatomic.h>
 #include <sys/time.h>
 
 #include <qthread/common.h>
 #include <qthread/qthread.h>
 
-#if (__STDC_VERSION__ >= 201112L) && (!defined(__STDC_NO_ATOMICS__))
-#define USE_C11_MEMORY_FENCE
-#include <stdatomic.h>
-#endif
-
-#ifdef USE_C11_MEMORY_FENCE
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Changing the name and marking as ready since I'm going to treat most of the thread sanitizer errors as individual issues and resolve them in individual PRs.

insertinterestingnamehere:
Reasoning behind this fix: the x86 memory model was making this work correctly there, but the data race is still undefined behavior WRT to the C standard. It appears that this particular race doesn't have correctness consequences, but it's still not desirable behavior. Switching to using atomic loads and stores basically tells all architectures to do what x86 is already doing here. Any specialized changes to improve performance on other architectures can be added later, but first let's eliminate the data race to standardize the behavior.

insertinterestingnamehere:
Converting back to a draft since https://github.com/sandialabs/qthreads/issues/187 seems more widespread.

insertinterestingnamehere:
Removed dependency on #185.

insertinterestingnamehere:
Okay, this one is good to go. There are still lingering sanitizer issues that show up in this file, but as best I can tell they come from #188 and #189, not #187. This PR just fixes #187.

insertinterestingnamehere:
Switching this one back to a draft again since I found some more errors connected to this part of the code that need to be resolved.

insertinterestingnamehere:
Okay, the lingering warnings about this section of code are in the `arbitrary_blocking_operation`, `external_fork`, `external_syncvar`, and `qthread_fork_precond` tests. Currently investigating why this patch is only a partial fix.

insertinterestingnamehere:
Okay, I think this one's good to go too, pending successful CI.

PR #206, opened 2023-12-05
Use Explicit Atomics In The Internal Spinlock

### LINKED ISSUE #190: Inadequate Atomic Guarantees In QTHREAD_FASTLOCK_LOCK
`QTHREAD_FASTLOCK_LOCK` and the related functions need to be rewritten to use atomic reads and writes instead of relying on the atomicity guarantees from x86. For example, there's currently a race condition between
https://github.com/sandialabs/qthreads/blob/d6ce514a70c65b74c5e04906615ec51c7f288e0f/src/feb.c#L986 and https://github.com/sandialabs/qthreads/blob/d6ce514a70c65b74c5e04906615ec51c7f288e0f/src/qthread.c#L652.

Note: this shows up readily in the hello_world_multi test.

insertinterestingnamehere:
Fixes #190.

### DIFF
--- a/include/qt_addrstat.h
+++ b/include/qt_addrstat.h
@@ -7,19 +7,19 @@
  * keeping track of the FEB status of an address. It expects a shepherd pointer
  * to use to find the right memory pool to use. */
 static QINLINE qthread_addrstat_t *qthread_addrstat_new(void)
-{                                      /*{{{ */
+{                                     /*{{{ */
     qthread_addrstat_t *ret = ALLOC_ADDRSTAT();
+    QTHREAD_FASTLOCK_INIT_PTR(&ret->lock);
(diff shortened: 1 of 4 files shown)


insertinterestingnamehere:
Still tracking some unusual bugs that are likely related to this so I'm converting it to a draft for now.

insertinterestingnamehere:
Okay, it looks like the other issues were resolved by other patches at this point. This one's good to go too. @janciesko can you review this one when you get a minute?

insertinterestingnamehere:
Throwing in a small fix for an intermittent race condition in this same section of code since it wouldn't make any sense without the context of this patch too.

janciesko (approved):
LGTM

PR #207, opened 2023-12-06
Fix Syncvar and FEB Lock/Unlock In Different Threads

### LINKED ISSUE #202: Unpaired Mutex Unlock In Syncvar
Found via a thread sanitizer warning in the `external_syncvar` test. There's a mutex unlock of an already unlocked mutex at https://github.com/sandialabs/qthreads/blob/d6ce514a70c65b74c5e04906615ec51c7f288e0f/src/syncvar.c#L312

insertinterestingnamehere:
Fixes #202.

As best I can tell, this function is only ever called in one place and the lock is managed by the caller there, so no unlock is needed. It's probably a leftover typo from some previous refactoring.

### DIFF
--- a/src/feb.c
+++ b/src/feb.c
@@ -61,6 +61,7 @@ typedef enum bt {
 } blocker_type;
 typedef struct {
     pthread_mutex_t lock;
+    pthread_cond_t condition;
     void           *a;
     void           *b;
     blocker_type    type;
@@ -230,7 +231,9 @@ static aligned_t qthread_feb_blocker_thread(void *arg)
             a->retval = qthread_empty(a->a);
             break;
     }
-    pthread_mutex_unlock(&(a->lock));
+    pthread_mutex_lock(&a->lock);
+    pthread_cond_signal(&a->condition);
+    pthread_mutex_unlock(&a->lock);
     return 0;
 }                                      /*}}} */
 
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Found the corresponding lock to this unlock. The issue wasn't that the unlock was unpaired it was that the unlock was getting called from a completely different thread.

The fix for that will be to either move the lock/unlock pairing either entirely into the called thread or entirely outside of it. I'm favoring locking outside of it here, but I'm not entirely sure of the performance implications of lock placement here so it'd be great to get a second opinion on that.

insertinterestingnamehere:
Hmm. Moving it inside may not work either under some circumstances even though that was working locally. Marking as WIP for now so I can decide what to do with it.

janciesko (approved):
LGTM

insertinterestingnamehere:
Alright! I sorted through what was going on here. This idiom was using a pthread mutex as a way to notify a waiting external thread that a blocked call had completed. Thread sanitizer was complaining since paired calls to lock/unlock from different threads aren't supposed to happen. Fortunately this is a fairly straightforward application of a condition variable so I just switched to using one of those.

insertinterestingnamehere:
I can't reproduce this locally. I'll set it down for now and circle back to it later.

janciesko:
What's the status of this?

insertinterestingnamehere:
This one's good to go now. It was causing some lingering issues in CI last time you looked at it, but I've fixed those since then.

PR #211, opened 2023-12-07
Explicit Atomics For IO

### LINKED ISSUE #196: Race Condition In IO Queue
There's a race condition in the IO queue where atomic loads and stores should be used instead of relying implicitly on the x86 memory model for consistency. The relevant lines are:
https://github.com/sandialabs/qthreads/blob/d6ce514a70c65b74c5e04906615ec51c7f288e0f/src/io.c#L386
https://github.com/sandialabs/qthreads/blob/d6ce514a70c65b74c5e04906615ec51c7f288e0f/src/io.c#L92

insertinterestingnamehere:
WIP. Hopefully will fix #196, but this patch appears to exchange a deadlock for the race condition instead. Opening this PR so CI can run and I can maybe get more info about what's going on.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -4,6 +4,7 @@
 
 /* System Headers */
 #include <qthread/qthread-int.h>       /* for uint64_t */
+#include <stdatomic.h>
 #include <stdio.h>                     /* for fprintf() */
 #include <stdlib.h>                    /* for abort() */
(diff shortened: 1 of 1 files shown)


janciesko:
LGTM

janciesko (approved):

PR #212, opened 2023-12-07
Atomics For Memory Pool

### LINKED ISSUE #189: Memory Pools Need Explicit Atomics
Our current memory pool functionality is relying on the x86 memory model for consistency and needs explicit atomics added to its caching functionality to avoid race conditions. For the moment, I'm seeing this because the sanitizers are raising errors about a data location having mixed atomic and non-atomic writes, but I'm fairly sure it represents an actual bug since the errors show one thread caching a freed node and another picking it up later.

An example of something that should be atomic and currently isn't: https://github.com/sandialabs/qthreads/blob/d6ce514a70c65b74c5e04906615ec51c7f288e0f/src/mpool.c#L415.

insertinterestingnamehere:
Fixes #189.

### DIFF
--- a/src/mpool.c
+++ b/src/mpool.c
@@ -61,19 +61,19 @@ struct qt_mpool_s {
 #else
     pthread_key_t                 threadlocal_cache;
 #endif
-    qt_mpool_threadlocal_cache_t *caches;  // for cleanup
+    qt_mpool_threadlocal_cache_t *_Atomic caches;  // for cleanup
 
     QTHREAD_FASTLOCK_TYPE         reuse_lock;
     void                         *reuse_pool;
 
     QTHREAD_FASTLOCK_TYPE         pool_lock;
     void                        **alloc_list;
-    size_t                        alloc_list_pos;
+    size_t _Atomic                alloc_list_pos;
 };
 
 typedef struct qt_mpool_cache_entry_s {
-    struct qt_mpool_cache_entry_s *next;
-    struct qt_mpool_cache_entry_s *block_tail;
+    struct qt_mpool_cache_entry_s *_Atomic next;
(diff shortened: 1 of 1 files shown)


src/mpool.c
```diff
@@ -409,35 +411,35 @@ void INTERNAL qt_mpool_free(qt_mpool pool,
...
     cache = tc->cache;
     cnt   = tc->count;
-    qthread_debug(MPOOL_DETAILS, "->cache:%p (bt:%p) cnt:%u\n", cache, cache ? cache->block_tail : NULL, (unsigned int)cnt);
+    qthread_debug(MPOOL_DETAILS, "->cache:%p (bt:%p) cnt:%u\n", cache, cache ? atomic_load_explicit(&cache->block_tail, memory_order_relaxed) : NULL, (unsigned int)cnt);
     if (cache) {
         assert(cnt != 0);
-        n->next       = cache;
-        n->block_tail = cache->block_tail; // cache is likely to be IN cache, so this won't be slow
+        atomic_store_explicit(&n->next, cache, memory_order_relaxed);
+        atomic_store_explicit(&n->block_tail, atomic_load_explicit(&cache->block_tail, memory_order_relaxed), memory_order_relaxed); // cache is likely to be IN cache, so this won't be slow
```
insertinterestingnamehere, line 418: I'm a little skeptical that this is enough synchronization here, but the main goal of this PR is to explicitly require the current behavior we're getting implicitly via the x86 memory model so that the behavior will be the same across architectures. More refactoring later on is probably a good idea for stuff like this.

janciesko: It should be ok as tc is pointing to TLS which is thread local.

janciesko (approved):
LGTM

PR #221, opened 2024-01-29
Remove Concurrent WriteFF Test

insertinterestingnamehere:
This test basically just tests that if you create a data race with our APIs it behaves correct WRT the underlying memory consistency guarantees from the architecture. As might be expected, the thread sanitizer doesn't like that, so I propose we at least remove this test.

I'd like to remove WriteFF entirely since the use-case is dubious (see https://github.com/sandialabs/qthreads/issues/220), but let's at least remove this bizarre test in the meantime.

### DIFF
--- a/test/basics/aligned_writeFF_basic.c
+++ b/test/basics/aligned_writeFF_basic.c
@@ -41,30 +41,6 @@ static aligned_t alignedWriteFF_iters(void *arg)
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


janciesko:
If we decide to remove writeFF, I am ok to remove this test.

olivier-snl:
Have we communicated with the Chapel team about writeFF?

Issue #225, opened 2024-02-02
Remove Our Own Implementations of Atomic Operations


insertinterestingnamehere:
We have a ton of internal implementations of various atomic operations that use either the old `__sync` gcc intrinsics or just use assembly directly. Given that we already require C11 atomics it'd be good to get rid of these things entirely to reduce the maintenance burden.

Issue #224, opened 2024-02-02
Explicit Atomics Needed In *incr Functions


insertinterestingnamehere:
This thread sanitizer failure only shows up when running the `qthread_fincr` and `qthread_dincr` tests on ARM. It should hopefully be straightforward to fix though.

For `fincr`, the issue is the non atomic load and then the atomic store to the same address at https://github.com/sandialabs/qthreads/blob/55ad590801b7188aa5f0e34a8946260a2726046c/include/qthread/qthread.h#L677 and then https://github.com/sandialabs/qthreads/blob/55ad590801b7188aa5f0e34a8946260a2726046c/include/qthread/qthread.h#L679.

The issue for dincr is similar. See https://github.com/sandialabs/qthreads/blob/55ad590801b7188aa5f0e34a8946260a2726046c/include/qthread/qthread.h#L833 and https://github.com/sandialabs/qthreads/blob/55ad590801b7188aa5f0e34a8946260a2726046c/include/qthread/qthread.h#L835.

PR #227, opened 2024-02-02
Explicit Atomics For Incr Fallback Implementations

insertinterestingnamehere:
Note: I think these functions should be totally rewritten using C11 atomics later (see #225), but this is just a minimal fix for #224.


insertinterestingnamehere:
Okay, the lingering CI failures appear resolved here now.

PR #231, opened 2024-02-06
Disable Thread Sanitizer Instrumentation for Context Swap

### LINKED ISSUE #219: Tests Segfault During Finalization With Thread Sanitizer Enabled
The following tests fail during finalization with thread sanitizer enabled (clang 17, nemesis scheduler, no topology detection, x86): `aligned_writeFF_basic`, `reinitialization`, `qthread_stackleft`, and `sinc_workers`. Interestingly, these tests do not fail when run by themselves or with a debugger. They also intermittently succeed. There appears to be some kind of segfault during finalization. At least with `aligned_writeFF_basic`, it successfully executes the test and then crashes instead of exiting normally. I'm investigating the behavior of the others to get more information.

insertinterestingnamehere:
Fixes #219.

Something about thread sanitizer's instrumentation doesn't like our context swap implementation. As documented in #219 there's a segfault during cleanup that happens because in some cases a worker thread fails to return from swapping the context to a qthread. This particular function isn't a source of concern for race conditions anyway so I'm just turning the sanitizer off locally.

### DIFF
--- a/src/fastcontext/context.c
+++ b/src/fastcontext/context.c
@@ -169,8 +169,22 @@ void INTERNAL qt_makectxt(uctxt_t *ucp,
 
 #endif /* ifdef NEEDPOWERMAKECONTEXT */
 
+// Macro for excluding a function from thread sanitizer.
+// Currently this is just used for qt_swapctxt.
+// Something about thread sanitizer's instrumentation is
+// incompatible with our context switching.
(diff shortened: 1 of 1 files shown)


janciesko:
Sounds good

janciesko (approved):

Issue #232, opened 2024-02-06
`qthread_disable_shepherd` Crashes Intermittently With Thread Sanitizer Enabled


insertinterestingnamehere:
Running on x86 with clang17, nemesis scheduler, no topology detection, and thread sanitizer enabled, the `qthread_disable_shepherd` test crashes intermittently. Unfortunately this only happens outside the debugger so more work is needed to find where exactly the segfault is happening.

insertinterestingnamehere:
Same thing was going on with `syncvar_prodcons`. Just added a commit that resolves this in #235.

insertinterestingnamehere:
Nevermind. It was just intermittent enough to look solved. The `syncvar_prodcons` intermittent failure is fixed though. Still investigating.

insertinterestingnamehere:
This hasn't been happening for a long while. Tentatively closing.

PR #235, opened 2024-02-07
Explicit Atomics in Syncvar Implementation

insertinterestingnamehere:
This one's more straightforward than the debate around `aligned_t` because syncvars were already treated as atomics internally. This was just a matter of making the loads and stores explicit.

I also made ARM take the same branch as x86 for these blocks of code since we're currently assuming lock-free atomics there too.


insertinterestingnamehere:
Now fixes #232 as well.

janciesko (approved):
LGTM

PR #243, opened 2024-02-20
Fix Intermittent Stackleft Crash With Thread Sanitizer

insertinterestingnamehere:
The intermittent failure with thread sanitizer can be eliminated by just getting rid of a little bit of the instrumentation (similar to #231). Given the slight incompatibility I already saw in #231 and the fact that this is not at all an invasive change I'm inclined to call this good enough.

While working on stackleft I also changed it to use `__builtin_frame_address(0)` instead of taking the address of a local variable to estimate the stack pointer since that seems like it'll be more reliable in the long run.


janciesko (approved):
LGTM

PR #246, opened 2024-02-22
Increase Default Stack Size

insertinterestingnamehere:
I've had to do this enough times while debugging that I'm proposing we just update the default. For some reason it seems to matter more on ARM systems, though I'm not sure why.


janciesko (approved):
LGTM

Issue #245, opened 2024-02-21
Race Conditions in Sherwood Threadqueue


insertinterestingnamehere:
The thread sanitizer is flagging several issues with the Sherwood threadqueue too. Here are a few of them:

Non-atomic write: https://github.com/sandialabs/qthreads/blob/a2d1dbadb80a161a4a287c1056348808b022ba04/src/threadqueues/sherwood_threadqueues.c#L322
Non-atomic read: https://github.com/sandialabs/qthreads/blob/a2d1dbadb80a161a4a287c1056348808b022ba04/src/threadqueues/sherwood_threadqueues.c#L845

Something non-atomic is still happening in the TRYLOCK macros/functions.
Non-atomic read at: https://github.com/sandialabs/qthreads/blob/a2d1dbadb80a161a4a287c1056348808b022ba04/src/threadqueues/sherwood_threadqueues.c#L846
Atomic write at: https://github.com/sandialabs/qthreads/blob/a2d1dbadb80a161a4a287c1056348808b022ba04/src/threadqueues/sherwood_threadqueues.c#L328

There are several more errors.

Issue #244, opened 2024-02-21
Race Conditions in Distrib Threadqueue


insertinterestingnamehere:
The thread sanitizer is flagging various issues in the distrib threadqueue. A few examples:

Non-atomic write: https://github.com/sandialabs/qthreads/blob/a2d1dbadb80a161a4a287c1056348808b022ba04/src/threadqueues/distrib_threadqueues.c#L207
Non-atomic read:
https://github.com/sandialabs/qthreads/blob/a2d1dbadb80a161a4a287c1056348808b022ba04/src/threadqueues/distrib_threadqueues.c#L265

Non-atomic read:
https://github.com/sandialabs/qthreads/blob/a2d1dbadb80a161a4a287c1056348808b022ba04/src/threadqueues/distrib_threadqueues.c#L216
Non-atomic write:
https://github.com/sandialabs/qthreads/blob/a2d1dbadb80a161a4a287c1056348808b022ba04/src/threadqueues/distrib_threadqueues.c#L374

There are several more.

PR #247, opened 2024-02-22
Fix Non-Atomic Load In Trylock

insertinterestingnamehere:
Some of the errors I was seeing in #244 and #245 came from this, but it doesn't have its own standalone issue.

### DIFF
--- a/include/qt_atomics.h
+++ b/include/qt_atomics.h
@@ -149,10 +149,9 @@ extern pthread_mutexattr_t _fastlock_attr;
 # define QTHREAD_TRYLOCK_INIT(x)     { (x).u = 0; }
 # define QTHREAD_TRYLOCK_INIT_PTR(x) { (x)->u = 0; }
 # define QTHREAD_TRYLOCK_LOCK(x)     { uint32_t val = qthread_incr(&(x)->s.users, 1); \
-                                       while (val != (x)->s.ticket) SPINLOCK_BODY(); \
-                                          THREAD_FENCE_MEM_ACQUIRE; /* spin waiting for my turn */ }
+                                       while (val != atomic_load_explicit((_Atomic uint32_t *)&(x)->s.ticket, memory_order_acquire)) SPINLOCK_BODY();}
 # define QTHREAD_TRYLOCK_UNLOCK(x)   do { COMPILER_FENCE; \
-					                      THREAD_FENCE_MEM_RELEASE; \
+					  THREAD_FENCE_MEM_RELEASE; \
                                           qthread_incr(&(x)->s.ticket, 1); /* allow next guy's turn */ \
                                         } while (0)
 # define QTHREAD_TRYLOCK_DESTROY(x)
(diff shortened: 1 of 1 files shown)


include/qt_atomics.h
```diff
@@ -149,18 +149,20 @@ extern pthread_mutexattr_t _fastlock_attr;
 # define QTHREAD_TRYLOCK_INIT(x)     { (x).u = 0; }
 # define QTHREAD_TRYLOCK_INIT_PTR(x) { (x)->u = 0; }
 # define QTHREAD_TRYLOCK_LOCK(x)     { uint32_t val = qthread_incr(&(x)->s.users, 1); \
-                                       while (val != (x)->s.ticket) SPINLOCK_BODY(); \
-                                          THREAD_FENCE_MEM_ACQUIRE; /* spin waiting for my turn */ }
+                                       while (val != atomic_load_explicit((_Atomic uint32_t *)&(x)->s.ticket, memory_order_acquire)) SPINLOCK_BODY();\
+                                       THREAD_FENCE_MEM_ACQUIRE;}
```
janciesko, line 153: Calling `SPINLOCK_BODY` seems superfluous in this case.

include/qt_atomics.h
```diff
@@ -149,18 +149,20 @@ extern pthread_mutexattr_t _fastlock_attr;
...
                                           qthread_incr(&(x)->s.ticket, 1); /* allow next guy's turn */ \
                                         } while (0)
 # define QTHREAD_TRYLOCK_DESTROY(x)
 # define QTHREAD_TRYLOCK_DESTROY_PTR(x)
 
 static inline int QTHREAD_TRYLOCK_TRY(qt_spin_trylock_t *x)
 {
-    qt_spin_trylock_t newcmp, cmp = *(x);
+    qt_spin_trylock_t newcmp, cmp;
+    uint64_t tmp = atomic_load_explicit((_Atomic uint64_t*)x, memory_order_acquire);
```
janciesko, line 164: I like the change. Likely this code used to work prior to this change as well due to the fence in the loop in L175.

include/qt_atomics.h
insertinterestingnamehere: Maybe? If `QTHREAD_OVERSUBSCRIPTION` is enabled, `SPINLOCK_BODY` can yield back to the threading runtime we're using. On the other hand I'm not sure if `QTHREAD_OVERSUBSCRIPTION` is even usable anymore so that may be a moot point.

Looking at this again though, I think the `THREAD_FENCE_MEM_ACQUIRE` can be removed here. Or, more precisely either that is needed or the `memory_order_acquire` in the loop condition. Not both.

include/qt_atomics.h
insertinterestingnamehere: Interesting point. I'm not sure the thread fence at 175 even matters because `qthread_cas` already has `acq_rel` consistency since it's implemented using the old gcc sync interface for atomics.

That aside, the main reasoning for this specific change is that we're already using an atomic access at this location in 174 and mixing atomic and non-atomic accesses is undefined behavior. I don't remember now why I added the acquire consistency here though. I'll go back and check.

include/qt_atomics.h
janciesko: I think we could remove `THREAD_FENCE_MEM_ACQUIRE`

include/qt_atomics.h
janciesko: Yes - I think we should be able to remove `THREAD_FENCE_MEM_ACQUIRE`. Let's see if we break something :)

janciesko (approved):
LGTM

PR #249, opened 2024-02-27
Thread Sanitizer Interop

insertinterestingnamehere:
This adds the API calls needed to help the thread sanitizer correctly carry its happens-before relationships across context swaps.

### DIFF
--- a/include/qt_qthread_struct.h
+++ b/include/qt_qthread_struct.h
@@ -3,6 +3,12 @@
 
 #include <stdatomic.h>
 
+#if defined(__has_feature)
+#if __has_feature(thread_sanitizer)
+#include <sanitizer/tsan_interface.h>
+#endif
+#endif
+
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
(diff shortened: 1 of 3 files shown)


insertinterestingnamehere:
Okay, doing another round of careful review of the fences in FEBs seems to have fixed the bugs I was trying to address here. On the other hand, this patch is still causing some mysterious issues. The interface I'm using is poorly documented so it's hard to say for sure if I'm using it correctly. I'm going to leave this open for a bit in case I need to come back to it, but for now I'm going to try to just move on without it.

insertinterestingnamehere:
Rebased to depend on #256 to avoid conflicts.

insertinterestingnamehere:
Nevermind, that was just a testing failure causing a false success. Marking this as a draft again for more debugging.

insertinterestingnamehere:
Okay, this still needs some restructuring for clarity, but it is actually working now.

insertinterestingnamehere:
Good enough. There's more to do to get thread sanitizer working correctly with our threadqueues, but this is a solid step forward.

PR #250, opened 2024-02-27
Adjust FEB Fences

insertinterestingnamehere:
Careful review appreciated for this one in particular.

This changes the implementation of FEBs so that a memory fence occurs even when the output value passed to the FEB routine is null. This matters because prior to this change, there's not necessarily a guaranteed fence implied by calls like `readFF` and the tests frequently use `readFF` to decide that it's safe to read some value that may be completely distinct from the FEB that's been marked as full.

### DIFF
--- a/test/basics/hello_world.c
+++ b/test/basics/hello_world.c
@@ -30,7 +30,7 @@ int main(int   argc,
     iprintf("%i shepherds...\n", qthread_num_shepherds());
     iprintf("  %i threads total\n", qthread_num_workers());
 
-    status = qthread_fork(greeter, (void*)123456789, &return_value);
+    status = qthread_fork(greeter, NULL, &return_value);
     assert(status == QTHREAD_SUCCESS);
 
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
@olivier-snl @janciesko This one's settled on my end now. I'd definitely still appreciate closer review on it though.

test/basics/hello_world.c
```diff
@@ -30,7 +30,7 @@ int main(int   argc,
     iprintf("%i shepherds...\n", qthread_num_shepherds());
     iprintf("  %i threads total\n", qthread_num_workers());
 
-    status = qthread_fork(greeter, (void*)123456789, &return_value);
+    status = qthread_fork(greeter, NULL, &return_value);
```
janciesko, line 33: Why this?

insertinterestingnamehere: That's just a readability fix. Sorry, for some reason I thought I'd included it with a different PR.

PR #257, opened 2024-04-12
Nemesis Cacheline Adjustment

insertinterestingnamehere:


insertinterestingnamehere:
Okay, testing this with the threadring benchmark on several different test machines shows that this does give a very small but consistent performance improvement so I'm inclined to go ahead with it.

janciesko:
Testing on M1

insertinterestingnamehere:
This either results in a very slight improvement or not change in performance on all systems tested. The only caveat is that the cacheline size needs to be set correctly, otherwise performance is wildly inconsistent with or without this patch.

PR #258, opened 2024-04-23
Add MVP `std::execution` qthreads scheduler with hello world test

msimberg:
Not ready for merging (if ever, perhaps this should be outside the repo?), just for discussion/testing.

Build with
```
./configure --enable-cxx-tests CXXFLAGS="-std=c++20 -I/path/to/stdexec/include -pthread" LDFLAGS="-lstdc++"
```

### DIFF
--- a/test/features/cxx_qt_stdexec.cpp
+++ b/test/features/cxx_qt_stdexec.cpp
@@ -1,3 +1,4 @@
+#include <qthread/qloop.h>
 #include <qthread/qthread.h>
 #include <stdexec/execution.hpp>
 #include <stdio.h>
@@ -64,7 +65,7 @@ struct qthreads_scheduler {
     }
 
     struct env {
-      qthreads_scheduler
+      friend qthreads_scheduler
       tag_invoke(stdexec::get_completion_scheduler_t<stdexec::set_value_t>,
                  env const &e) noexcept {
         return {};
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Yah, I'm fine with including this in the main repo. I think the open questions are just how we're doing our branching and what the default C++ build config should be. It's a useful point of reference though so it'd be great to include it in some form.

test/features/cxx_qt_stdexec.cpp
```diff
@@ -0,0 +1,103 @@
...
+      auto *os = static_cast<operation_state *>(arg);
+      printf("Hello from qthreads in task! id = %i\n", qthread_id());
+      stdexec::set_value(std::move(os->receiver));
+      return 0;
+    }
+
+    friend void tag_invoke(stdexec::start_t, operation_state &os) noexcept {
+      aligned_t ret = 0;
+      int r = qthread_fork(&task, &os, &ret);
+      qthread_readFF(NULL, &ret);
```
insertinterestingnamehere, line 39: Is this the right place to put the `readFF` call? I think this is roughly the same as running `task` synchronously since the calling thread waits on the `readFF` call.

msimberg: I was wondering the same.

Just to clarify, the `sync_wait` sender consumer sets up a condition variable that it starts waiting for. The condition variable is signaled by the `set_value` call that is called inside `task`. So synchronization is taken care of without the `qthread_readFF`, but this only guarantees synchronization until user-supplied callables have finished running. The qthread task obviously hasn't fully finished running when `sync_wait` returns, and I had assumed this would be ok to do.

It is possible to customize `sync_wait` and/or provide a fiber-specific `sync_wait` that yields the fiber instead of the underlying OS-thread. In this particular example it doesn't matter since `sync_wait` is called on `main`, and in general you want to avoid calling `sync_wait` except at the top level. I think if one sets up a customized `sync_wait` that one could use qthreads' own synchronization mechanisms, like `readFF`.

janciesko:
I would like to propose to continue the development of the stdexec ult examples here https://github.com/janciesko/kokkos-stdexec-examples. Once we're happy, we should move the repo to Kokkos.org.

PR #260, opened 2024-05-28
Gcc 14 in CI

insertinterestingnamehere:


insertinterestingnamehere:
Looks like the toolchain test builds aren't up yet for this release (https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test). I'll just leave the PR as-is and revisit it when those packages become available.

janciesko:
Thanks, this makes sense

janciesko (approved):

insertinterestingnamehere:
Okay, it turns out gcc14 is available in the ubuntu universe repository for 24.04 so I just switched the tests to using that build.
While I'm here I'm also doing slight version bumps for ACFL and icx.

PR #266, opened 2024-08-08
Increase MUSL CI Timeout Limit

insertinterestingnamehere:


insertinterestingnamehere:
Interesting, dramatically increasing the timeout limit still only partially eliminates the problems here. I'll try rerunning the build a few times to get a better sense for how often the errors are happening now and then decide where to go from there.

insertinterestingnamehere:
Yah, this only marginally improves things, so I'm going to call the hanging a proper bug and create an issue for it instead.

PR #269, opened 2024-08-27
Adjust Fences on X86 to Correct Performance Regression

insertinterestingnamehere:


insertinterestingnamehere:
How odd. Adding a test to make sure this was okay seems to have unearthed a completely separate bug. The new test fails before and after the change to the fences so that's not what's causing it. I'll split the new test off into a separate PR.

PR #273, opened 2024-09-11
Drop Old Architectures

insertinterestingnamehere:


insertinterestingnamehere:
Leaving 32 bit architectures alone in this pass since that's more complicated. This at least partially solves https://github.com/sandialabs/qthreads/issues/233 though.

PR #277, opened 2024-09-16
Drop Various Rare OSs and Remove Unnecessary Config Checks

insertinterestingnamehere:
Depends on https://github.com/sandialabs/qthreads/pull/274 since they modify very many of the same files.


insertinterestingnamehere:
Looks like there's still an OSX-specific error to chase down, but other than that this is a good stopping point for this PR. I basically purged whatever could easily be removed from the configure directory. Further removals can be dealt with in separate PRs.

insertinterestingnamehere:
Okay, OSX issue resolved.

PR #286, opened 2024-09-23
Use C11 Atomics Exclusively

insertinterestingnamehere:
Fixes https://github.com/sandialabs/qthreads/issues/234 and https://github.com/sandialabs/qthreads/issues/204.


src/cacheline.c
```diff
@@ -8,6 +8,8 @@
 #include <qthread/cacheline.h>
 #include <qthread/common.h>
 
+#define HAVE_GCC_INLINE_ASSEMBLY
```
janciesko, line 11: Is this a fly-by (unrelated) change?

insertinterestingnamehere: I opted to not remove that macro from that specific file so I just defined it at the top instead. The whole logic there needs an overhaul. I'm fine going back and cutting it out there too if that'd be preferable.

janciesko:
LGTM

PR #296, opened 2024-10-03
Test nvc in CI

insertinterestingnamehere:


insertinterestingnamehere:
Looks like nvc doesn't doesn't (yet) support the gcc flags that automake is trying to use to generate dependency files. I think this is similar to what's discussed at: https://forums.developer.nvidia.com/t/issue-with-header-file-dependency-file-d-generation/221608. The fix isn't immediately obvious to me, but I'll leave this PR open for now.

insertinterestingnamehere:
Alright, here's the plan: the dependency flag issue only applies to our assembly file so we're going to test nvc with our fast context swaps disabled until we can get the build system to play nice with its restricted flag options. That way this doesn't get blocked.

Some of the nvc warnings appear to be buggy, but I also fixed the ones I could as a part of this PR too.

insertinterestingnamehere:
Github actions appears to be having an outage right now, so I'll come back and merge this in a bit once they get that working again so we can confirm that I didn't somehow break an existing supported config.

PR #297, opened 2024-10-08
Increase Various OS and Compiler Versions in CI

insertinterestingnamehere:


insertinterestingnamehere:
Alright, this looks good now. Worth noting: we're now not testing clang versions 13 and older since that's what's easily available on the newer Ubuntu LTS version. They don't break often so they probably will continue to work fine, but we may want to consider not labeling them as "supported" anymore. We still support all the way back to gcc 9, so this should be fine for people on old systems.

PR #314, opened 2024-11-07
Minor Performance Regression Fixes

insertinterestingnamehere:
These are some minor performance fixes to help fix regressions downstream

In particular, this removes an unnecessary thread fence in the IO code, and updates the MACHINE_FENCE macro to use a stronger type of memory fence (since this somehow results in better performance in practice).

The trylock patch is just a simplification that I wrote while attempting to debug the performance regressions. That should be performance neutral.

### DIFF
--- a/include/qthread/qthread.h
+++ b/include/qthread/qthread.h
@@ -87,7 +87,7 @@ using std::memory_order_relaxed;
 
 #include "macros.h"
 
-#define MACHINE_FENCE atomic_thread_fence(memory_order_acq_rel);
+#define MACHINE_FENCE atomic_thread_fence(memory_order_seq_cst);
 
 #if QTHREAD_ASSEMBLY_ARCH == QTHREAD_AMD64
 #define QTHREAD_SWAPS_IMPLY_ACQ_REL_FENCES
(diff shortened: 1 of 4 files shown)


include/qthread/qthread.h
```diff
@@ -87,7 +87,7 @@ using std::memory_order_relaxed;
 
 #include "macros.h"
 
-#define MACHINE_FENCE atomic_thread_fence(memory_order_acq_rel);
```
janciesko, line 90: Is change is needed for correctness or for performance?

insertinterestingnamehere: Surprisingly, performance. I'm still confused as to how/why this is happening, but for the chapel benchmark I'm using to track the regression there's a modest performance improvement from using the sequentially consistent fences instead of just acquire release consistent. This is consistent with what the old assembly based version was doing prior to https://github.com/sandialabs/qthreads/commit/b122234e164be9bc6bc1cd3d45673e53c7f22f95 anyway, but it's still weird.

insertinterestingnamehere: I'm not aware of anything in our codebase that ought to actually require sequential consistency for correctness' sake. The fact that sequential consistency is beneficial here seems to indicate that there's something suboptimal going on internally. This patch at least corrects the downstream performance regression though.

PR #315, opened 2024-12-17
Additional Performance Fixes

insertinterestingnamehere:
This reverts one of the commits from #314 since it ended up being more harmful than helpful downstream. The Chapel array init/deinit benchmark seems to be very hardware dependent and we don't want to eliminate the big improvements in other applications from the 1.21 release.


insertinterestingnamehere:
CI failures are unrelated. Hopefully this is good enough to help downstream.

PR #323, opened 2025-01-07
Switch CI Builds To CMake

insertinterestingnamehere:


insertinterestingnamehere:
Hmm. In theory this is done, but there's still an odd very intermittent failure that shows up only when building with CMake. There's not an obvious cause in the differences between the builds, so I'm still debugging.

insertinterestingnamehere:
Oooookay, dropping the optimization level from 3 to 2 seems to have mitigated this so much that it's become vanishingly rare (once every few thousand runs). Given that the queue test was turned off until I reinstated it while reorganizing some files and that this only happens in the (seldom used) sherwood/binders config I'm just going to document the error and move on.

PR #333, opened 2025-01-14
Atomic Queuelengths

insertinterestingnamehere:
This is an experimental PR that provides an alternate fix for https://github.com/sandialabs/qthreads/issues/324.


insertinterestingnamehere:
Alright, looks like this isn't enough to fix that discrepancy. The code as it was before this change looks wrong though, so I think it may be best to leave the test disabled but also still merge this. I'll consider how to proceed.

PR #336, opened 2025-01-22
Support Older CMake Versions

insertinterestingnamehere:
Still needs some testing.


insertinterestingnamehere:
CI failures appear to all be incarnations of https://github.com/sandialabs/qthreads/issues/288, so completely unrelated. I'm still testing the install behavior with older versions to make sure I have it at least right-ish though.

PR #340, opened 2025-02-18
Version Bump For 1.22 Release Candidate

insertinterestingnamehere:


insertinterestingnamehere:
Including a very minor CI fix for an error that just showed up. They're unrelated, I just don't want to do two separate PRs since these changes are both miniscule.

PR #355, opened 2025-04-07
Only Default To Fastcontext When It's Known To Work

insertinterestingnamehere:
Inspired by a bug on a "ppc64le" machine. It turned out that the old build system was using the fallback implementation on that platform which was hiding the fact that our context swap code doesn't actually work there. We're only nominally committed to supporting any kind of powerpc platform whatsoever so this just makes the new build system fall back to the system swapcontext on that platform. We may just fix the little-endian bug or just drop support for that platform later, but this at least maintains the baseline of working on that platform for now.


insertinterestingnamehere:
Okay, the CI failures are unrelated. They appear to be caused by some kind of change upstream. I'm debugging them elsewhere. 
This is tentatively good to go so I'll squash, wait for another round of CI, then merge.

bradcray:
Thanks for the quick turnaround on this, Ian!

PR #384, opened 2026-05-04
Consolidate CI Builds

insertinterestingnamehere:
We've been running each possible build configuration in a separate container. This change reduces the total number of package downloads that have to happen by combining different all the different build config tests that need to run in the same build environment. This will hopefully make the test suite a bit faster and more robust to download delays.


insertinterestingnamehere:
Failure is unrelated. Merging.

Issue #377, opened 2026-01-27
cxx_qt_loop_balance Fails With Latest nvc


insertinterestingnamehere:
First observed in https://github.com/sandialabs/qthreads/pull/374 with nvc 25.11. Likely a bug upstream, but a workaround is needed regardless.

insertinterestingnamehere:
Found another example of this: https://app.circleci.com/pipelines/github/sandialabs/qthreads/144/workflows/1e108a76-0ad6-4de5-b0d2-79292148d6aa/jobs/11057. Apparently it's just very intermittent.

insertinterestingnamehere:
Another example: https://github.com/sandialabs/qthreads/actions/runs/32284716464/job/96171627353?pr=381

insertinterestingnamehere:
Confirmed present with nvc 26.5

insertinterestingnamehere:
Successfully reproduced locally. If I insert a print statement before https://github.com/sandialabs/qthreads/blob/ca50f11d10372b72668ef37c0751487e9bd9e687/src/qloop.c#L249, the bug disappears. I can't get it to appear when run inside gdb. Working on getting a core dump instead. Note: last time we had a bug like this it seemed to arise from the mixed use of atomics between the c11 atomic routines and the older __sync intrinsics. In this case though, it's fairly easy to swap everything over to use c11 atomics but that does not appear to fix it.

insertinterestingnamehere:
A few example backtraces below. The thread 1 backtrace is the same each time, and it happens right at the context swap. Presumably that's where the crash actually happens. Cause still unclear.

```
Thread 2:
#0  in qthread_worker@plt () from qthreads/build/src/libqthread.so.1
#1  in qt_scheduler_get_thread () at qthreads/src/threadqueues/distrib_threadqueues.c:412
#2  in qthread_master () at qthreads/src/qthread.c:360
#3  in start_thread () from libc.so.6
#4  in thread_start () from libc.so.6

Thread 1:
#0  in qt_setmctxt () from qthreads/build/src/libqthread.so.1
#1  in qthread_back_to_master2 () at qthreads/src/qthread.c:2193
#2  qthread_wrapper () at qthreads/src/qthread.c:1645
#3  in ?? () at qthreads/src/qthread.c:1500 from qthreads/build/src/libqthread.so.1
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```

```
Thread 2:
[...]

PR #403, opened 2026-09-18
Simplify arm64 Context Swap

insertinterestingnamehere:
While debugging something else I noticed this code was doing a bunch of unnecessary work.

WIP, creating this early to check the CI results quickly.


insertinterestingnamehere:
Oooookay. Weirdly, poking at this code seems to have fixed #377. There's another minor simplification I want to do in this area, but this seems to be working great.

insertinterestingnamehere:
Scratch that. Eliminating the additional unused struct member revives the bug in #377. That's definitely wrong, so I'm leaving that one open for now. Going to see if I can get any additional debugging info out of that variation though.

